PR #87, opened 2014-08-22
Fix byte compilation in utop-choose

lunaryorn:
Fix the flawed `cond` branch to correctly signal an error for unsupported modes, and wrap `utop-choose` in `eval-and-compile` to make it available at macro expansion time.

This whole `utop-choose` thing looks really scary.  What problem is it even supposed to solve?

### DIFF
--- a/src/top/utop.el
+++ b/src/top/utop.el
@@ -759,7 +759,8 @@ when byte-compiling."
     (intern (concat "typerex-" symbol)))
    ((require 'caml nil t)
     (intern (concat "caml-" symbol)))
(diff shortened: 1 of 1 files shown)


mads379:
I've been playing around with this for a bit because the modification of tuareg-mode variables was giving me issues when using org-babel. Please have a look this change-set https://github.com/mads379/utop/compare/refactor-major-mode-use?expand=1

I haven't opened a PR yet as I'm not sure how to handle `utop-skip-blank-and-comments`, `utop-skip-to-end-of-phrase` and `utop-discover-phrase`. Currently I've bound them to the `tuareg-` version of each but that of course won't work for other modes. Any hints on how to get around this?

diml:
I'm closing this PR as the issue should be fixed by #124

PR #124, opened 2015-01-05
Refactor major mode use

mads379:
These changes were sparked by the discussion on #87.

Additionally I was hitting a bug when using the babel feature in org-mode. The previous implementation would mess with the `*-interactive-buffer-name` variable which stopped babel from being able to evaluate my OCaml code in my code blocks.

Please note that this is a breaking-change for emacs users as they will need to change their configuration given that `utop-setup-ocaml-buffer` has been deleted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ utop also ships with a minor mode that has the following key-bindings
 
 | key-binding | function          | Description                  |
 |-------------|-------------------|------------------------------|
-| C-c C-z     | utop              | Start a utop buffer          |
+| C-c C-s     | utop              | Start a utop buffer          |
 | C-x C-e     | utop-eval-phrase  | Evaluate the current phrase  |
 | C-x C-r     | utop-eval-region  | Evaluate the selected region |
 | C-c C-b     | utop-eval-buffer  | Evaluate the current buffer  |
@@ -158,9 +158,6 @@ e.g. typerex, then you will need the following configuration instead
 
 ```scheme
 (autoload 'utop-minor-mode "utop" "Minor mode for utop" t)
-(setq utop-skip-blank-and-comments 'typerex-skip-blank-and-comments)
-(setq utop-skip-to-end-of-phrase 'typerex-skip-to-end-of-phrase)
-(setq utop-discover-phrase 'typerex-discover-phrase)
 (add-hook 'typerex-mode-hook 'utop-minor-mode)
 ```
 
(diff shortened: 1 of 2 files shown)


diml:
Thanks for working on this!

I think to make the transition easier we should keep an `utop-setup-ocaml-buffer` function that prints an error message.

src/top/utop.el
```diff
@@ -207,13 +207,30 @@ before the end of prompt.")
...
 
-(defvar utop-ocaml-preprocessor nil
-  "Name of preprocesor. Currently supported camlp4o, camlp4r.
-Useful as file variable.")
+(make-variable-buffer-local
+ (defvar utop-ocaml-preprocessor nil
+   "Name of preprocesor. Currently supported camlp4o, camlp4r.
+Useful as file variable."))
+
+(defvar utop-skip-blank-and-comments 'tuareg-skip-blank-and-comments
```
diml: It'd be nice if the default value was a function that dispatch according to the current mode to handle tuareg, caml and typerex. It doesn't seem too bad and make the transition easier for people who use either caml or typerex. Same for the next two variables.

README.md
```diff
@@ -135,21 +135,38 @@ you can copy-paste this code into you `~/.emacs` file:
...
 
-Integration with the tuareg/typerex mode
-----------------------------------------
+utop also ships with a minor mode that has the following key-bindings
 
-You can replace the default toplevel used by the tuareg or typerex
-mode by utop, for that add the following lines to your `~/.emacs` file:
+| key-binding | function          | Description                  |
+|-------------|-------------------|------------------------------|
+| C-c C-z     | utop              | Start a utop buffer          |
```
diml: Why not use `C-c C-s` like tuareg?

mads379:
@diml Alright, please have a look at my latest commit :)

diml:
Looks good, thanks!

mads379:
Awesome, thanks!

spacebat:
I just wonder, given that https://github.com/realworldocaml/book/wiki/Installation-Instructions includes a call to utop-setup-ocaml-buffer, whether the deprecation error message should give a little more information. Is the new equivalent of (utop-setup-ocaml-buffer) a no-op; ie should I just delete it or replace it with something else?

diml:
@spacebat you are right, I updated the error message in a721d130826ffc820f0591806f12ba28ff33ca1c. Feel free to submit a pull request if you think it can be improved.

PR #129, opened 2015-03-09
Adapt for 4.03.0

Drup:
Keep uptodate with trunk. I didn't fix the deprecation warnings.

### DIFF
--- a/src/lib/uTop_main.cppo.ml
+++ b/src/lib/uTop_main.cppo.ml
@@ -283,6 +283,9 @@ let rec map_items unwrap wrap items =
       | Outcometree.Osig_modtype (name, _)
       | Outcometree.Osig_value (name, _, _) ->
         (name, Outcometree.Orec_not)
+#if OCAML_VERSION >= (4, 03, 0)
+      | Outcometree.Osig_ellipsis -> assert false
+#endif
     in
     let keep = name = "" || name.[0] <> '_' in
     if keep then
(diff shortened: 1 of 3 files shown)


src/lib/uTop_main.cppo.ml
```diff
@@ -283,6 +283,9 @@ let rec map_items unwrap wrap items =
       | Outcometree.Osig_modtype (name, _)
       | Outcometree.Osig_value (name, _, _) ->
         (name, Outcometree.Orec_not)
+#if OCAML_VERSION >= (4, 03, 0)
+      | Outcometree.Osig_ellipsis -> assert false
+#endif
```
Drup: Beware here. I don't know how to fix since I mostly don't know utop's internals.

src/lib/uTop_main.cppo.ml
```diff
@@ -322,8 +325,11 @@ let rec map_items unwrap wrap items =
           | Outcometree.Osig_exception _
 #endif
           | Outcometree.Osig_modtype _
-          | Outcometree.Osig_value _ ->
-            items
+          | Outcometree.Osig_value _
+#if OCAML_VERSION >= (4, 03, 0)
+          | Outcometree.Osig_ellipsis
+#endif
+           -> items
```
Drup: I think this is right, not completely sure.

src/lib/uTop_main.cppo.ml
whitequark: Should be `| Outcometree.Osig_ellipsis -> Outcometree.Osig_ellipsis`.

src/lib/uTop_main.cppo.ml
whitequark: Looks good.

src/lib/uTop_main.cppo.ml
Drup: It's supposed to return a pair `(name, something)` and there is no name to get.

whitequark: Oops. I think it should return "" as name--see the line just below.

Drup:
I think It can be closed now. :)

PR #145, opened 2015-11-05
Fix non-exist function

syohex:
Emacs does not have 'replace-in-string' function. You can see this by byte-compiling.

```
In end of data:
utop.el:1258:1:Warning: the function `replace-in-string' is not known to be defined.
```


diml:
Thanks!

PR #169, opened 2016-06-25
Stashing

chrismamo1:
I frequently find myself putting a lot of work into a project with UTop and wishing to save that work without the tedium of manually rooting through the terminal history. This PR includes a `#utop_stash "filename"` directive which writes all of the valid toplevel expressions executed in the active session to a file. In order to accomplish this, it maintains a separate history which is never backed onto a file unless requested using the directive above, and contains only the toplevel expressions from the current session which compiled without errors.

### DIFF
--- a/src/lib/uTop.cppo.ml
+++ b/src/lib/uTop.cppo.ml
@@ -636,8 +636,7 @@ let () =
       (fun fname ->
         let _ :: entries = LTerm_history.contents stashable_session_history in
         (* getting and then reversing the entries instead of using
-         * [LTerm_history.save] because the latter has to be integrated
-         * into an LWT program flow. *)
+           [LTerm_history.save] because the latter escapes newline characters *)
         let () =
           Printf.printf
             "Stashing %d entries in %s... "
@@ -648,12 +647,10 @@ let () =
         try
           let oc = open_out fname in
           try
-            let ls = ref entries in
-            for i = 1 to List.length entries do
-              let e :: _ = !ls in
-              let () = output_string oc (e ^ "\n") in
-              ls := List.tl !ls;
-            done;
+            List.iter
+              (fun e ->
+                output_string oc (e ^ "\n"))
+              entries;
             close_out oc;
             Printf.printf "Done.\n";
           with exn ->
--- a/src/lib/uTop.cppo.ml
+++ b/src/lib/uTop.cppo.ml
@@ -27,6 +27,11 @@ let history = LTerm_history.create []
 let history_file_name = ref (Some (Filename.concat LTerm_resources.home ".utop-history"))
 let history_file_max_size = ref None
 let history_file_max_entries = ref None
+let stashable_session_history =
+  (LTerm_history.create
+    ~max_size:max_int
+    ~max_entries:max_int
+    [])
 
 (* +-----------------------------------------------------------------+
    | Hooks                                                           |
@@ -532,6 +537,7 @@ utop defines the following directives:
 
 #utop_bindings   : list all the current key bindings
 #utop_macro      : display the currently recorded macro
+#utop_stash      : store all the valid commands from your current session in a file
 #topfind_log     : display messages recorded from findlib since the beginning of the session
 #topfind_verbose : enable/disable topfind verbosity
 
@@ -623,6 +629,37 @@ let () =
   Hashtbl.add Toploop.directive_table "pwd"
     (Toploop.Directive_none
        (fun () -> print_endline (Sys.getcwd ())))
+
+let () =
+  Hashtbl.add Toploop.directive_table "utop_stash"
+    (Toploop.Directive_string
+      (fun fname ->
+        let _ :: entries = LTerm_history.contents stashable_session_history in
+        (* getting and then reversing the entries instead of using
+         * [LTerm_history.save] because the latter has to be integrated
+         * into an LWT program flow. *)
+        let () =
+          Printf.printf
+            "Stashing %d entries in %s... "
+            (List.length entries)
+            fname
+        in
+        let entries = List.rev entries in
+        try
+          let oc = open_out fname in
+          try
+            let ls = ref entries in
+            for i = 1 to List.length entries do
+              let e :: _ = !ls in
+              let () = output_string oc (e ^ "\n") in
+              ls := List.tl !ls;
+            done;
+            close_out oc;
+            Printf.printf "Done.\n";
+          with exn ->
+            close_out oc;
+            Printf.printf "Done.\n";
+        with exn -> Printf.printf "Error with file %s.\n" fname))
 
 (* +-----------------------------------------------------------------+
    | Camlp4                                                          |
--- a/src/lib/uTop.cppo.ml
+++ b/src/lib/uTop.cppo.ml
@@ -640,7 +640,7 @@ let () =
         let () =
           Printf.printf
             "Stashing %d entries in %s... "
-            (List.length entries)
+            (List.length entries / 2) (* because half are comments *)
             fname
         in
         let entries = List.rev entries in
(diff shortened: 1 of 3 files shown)


src/lib/uTop.cppo.ml
```diff
@@ -624,6 +630,37 @@ let () =
...
        (fun () -> print_endline (Sys.getcwd ())))
 
+let () =
+  Hashtbl.add Toploop.directive_table "utop_stash"
+    (Toploop.Directive_string
+      (fun fname ->
+        let _ :: entries = LTerm_history.contents stashable_session_history in
+        (* getting and then reversing the entries instead of using
+         * [LTerm_history.save] because the latter has to be integrated
+         * into an LWT program flow. *)
```
diml: This is not a good reason as you can use `Lwt_main.run` here. Other directives are calling it. But I imagine you don't want to escape newlines as `LTerm_history.save` does. Can you update the comment?

src/lib/uTop.cppo.ml
```diff
@@ -624,6 +630,37 @@ let () =
...
+            "Stashing %d entries in %s... "
+            (List.length entries)
+            fname
+        in
+        let entries = List.rev entries in
+        try
+          let oc = open_out fname in
+          try
+            let ls = ref entries in
+            for i = 1 to List.length entries do
```
diml: Why not `List.iter`?

src/lib/uTop.mli
```diff
@@ -210,6 +210,15 @@ val history_file_max_entries : int option ref
...
       used. *)
 
+val stashable_session_history : LTerm_history.t
+  (** A history consisting only of expressions successfully evaluated
+   * during the current session. Because stashing is supposed to
+   * produce a valid OCaml file which will behave roughly the same as
+   * the console, it is best if this history never gets truncated, so
+   * its maximum size is set at [max_int]. While this will certainly
+   * lead to a slight memory leaking problem, UTop sessions are
+   * rarely long enough to make it a serious issue. *)
```
diml: This style of comment is not used in the utop code, could you remove the leading stars?

src/lib/uTop_main.cppo.ml
```diff
@@ -183,6 +183,11 @@ class read_phrase ~term = object(self)
           let result = parse_and_check input eos_is_error in
           return_value <- Some result;
           LTerm_history.add UTop.history input;
+          ignore(match result with
+            | UTop.Value _, _ ->
+                LTerm_history.add UTop.history input;
```
diml: This is not needed, right?

diml:
Thanks, that's a useful feature.

Something similar I'd like to have at some point is the ability to save the whole session, including inputs and responses, with a simplified prompt. Given that copy&pasting from the terminal into an email doesn't always work well, it would make it easier to send a piece of utop session as an example.

chrismamo1:
Thanks for the feedback, I'll update the code and submit a patch sometime today.

src/lib/uTop.cppo.ml
chrismamo1: @diml I apologize, I'm not too familiar with collaborating on projects that use LWT, and I just assumed that having more than one `Lwt_main.run` in an application was poor form. You are right, however, that escaped newlines are not the desired behaviour here.

src/lib/uTop.cppo.ml
chrismamo1: Noted, changing right now.

src/lib/uTop.mli
chrismamo1: Changed, though I'm not sure what the column limit is in uTop code. I'm familiar with the 80 column rule, but most of the existing comments seem to stop well before then.

src/lib/uTop_main.cppo.ml
chrismamo1: @diml The stashing feature maintains a separate history (it only makes sense to stash lines which compiled without incident). As I understand the code here, the `result` against which I am matching on the highlighted lines will only match with the highlighted pattern if its expression compiled successfully.

chrismamo1:
I have a couple questions regarding the precise behaviour of this feature, and would like the input of the maintainers:

1. Should stashing also reset the stashable history, so that you only stash lines which were run since the last time you ran `#utop_stash ...`? I'm inclined to say no, because new commands could depend on the ones previously stashed.
2. @diml brought up stashing an entire session including responses. I was planning on doing this, with the responses appearing in comments in the lines following their corresponding commands, e.g.:

        let x = 37;;
        (* val x : int = 37 *)
        x + 12;;
        (* - : int = 49 *)

    Would this work for you? Looking at the code right now it seems relatively quick to write up, so I might just submit a new patch sometime today.
3. I would very much like it if stashed files could be compiled and executed immediately, with no modifications. Perhaps the stashing function could automatically produce a shebang directive and necessary ocamlscript headers? This could turn out to be a fair amount of work, since it would require processing all of the `#require` directives from the current session, but I'm relatively free for the next week.

hcarty:
The ocamlscript idea sounds really cool! Does ocamlscript work with ppx syntaxes? I think it may use camlp4 internally, but it's been a while since I've used it.

src/lib/uTop_main.cppo.ml
chrismamo1: @diml I just noticed what you were talking about, and no it is not necessary to save the normal history again within the `ignore`.

chrismamo1:
@hcarty

> Does ocamlscript work with ppx syntaxes?

I have no idea, my experience with ocamlscript is extremely limited, but what I know about it makes it seem like an appropriate tool.

chrismamo1:
Please pardon all the extraneous messages that just popped up, I'm still not completely solid with rebasing.

diml:
1. I'm fine with not resetting the stashable history. Especially if we want to produce a working script, it makes sense to keep the whole session

2. I was thinking of something like this:

```
# let x = 37;;
val x : int = 37
# x + 12;;
- : int = 49
# x ^ "foo";;
Error: This expression has type int but an expression was expected of type
         bytes
```

i.e. close to what you see in the terminal. It should include errors as well. But that can be a separate command with a separate history.

3. That'd be very nice indeed. Maybe an easier goal would be to produce an ocaml (or utop) script. Then you could just output `#require`  and other directives as it

chrismamo1:
After the latest commit, stashing output appears as:

```ocaml
let x = 37;;
(* val x : int = 37 *)

x + 12;;
(* - : int = 49 *)

(* x ^ "foo";; *)
(* ESC[1;31mErrorESC[0m: This expression has type int
       but an expression was expected of type 
       string *)

type some_type = Foo | Bar | Foobar;;
(* type some_type = Foo | Bar | Foobar *)

let some_function = function
         | Foo -> 0
         | Bar -> 1;;
(* val some_function : some_type -> int = <fun> *)

(* x = ;; *)
(* Error: Syntax error *)
```

Unfortunately, I can't for the life of me figure out how to get warnings properly printed.

chrismamo1:
The biggest problem I can see with the current output is that it is trying to print the colouring information for the user's terminal, which is annoying, but could be useful (`cat stash.ml` could be enhanced by this). I could look into fixing this?

diml:
I think we should strip the escaping sequences. There should be a way to not print them, but stripping them should be easy, we just need to drop everything between a `'\027'` and a `'m'`.

BTW, I'd still be interested in a way to output something that looks like a toplevel session. Maybe instead of using a `LTerm_history.t`, we should use a list of:

```ocaml
type t = Input of string | Response of string
```

then different stashing function could print things differently.

For warnings, initially utop used a hack to capture them and format them. Since then, the compiler got new hooks in  the `Location` module for printing warnings so this could be use to capture them.

diml:
Superseeded by #199

PR #172, opened 2016-07-04
Handle the `Invalid_package_type` syntax error

Octachron:
When using OCaml 4.03, entering in utop

```OCaml
(module List: sig end);;
```
makes utop crashes with
```
Fatal error: exception File "src/lib/uTop.cppo.ml", line 246, characters 31-36: Pattern matching failed .   
```
The crash appears to stem from the new `Invalid_package_type` syntax error introduced in Ocaml 4.03 (for [PR#7097](http://caml.inria.fr/mantis/view.php?id=7097), [change entry](https://github.com/ocaml/ocaml/blob/trunk/Changes#L363)) . This pull request simply adds this case in the relevant pattern matching to fix this fatal error.

### DIFF
--- a/src/lib/uTop.cppo.ml
+++ b/src/lib/uTop.cppo.ml
@@ -268,6 +268,11 @@ let parse_default parse str eos_is_error =
           Error ([mkloc loc],
                  Printf.sprintf "Error: broken invariant in parsetree: %s" s)
 #endif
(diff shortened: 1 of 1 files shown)


diml:
This was fixed by d888ee5ba9bc17e8f965e64e65de1cc813d3aac7

PR #182, opened 2016-10-07
Add source directive for accurate error line numbers

rleonid:


diml:
thanks, in fact oasis should probably do this...

PR #183, opened 2016-10-07
Constrain the oasis dependency

rleonid:


diml:
you should be able to write this in the opam file:

```
conflicts: [ "oasis" {= "0.4.7"} ]
```

it's better as the fix is likely to go in the next version of oasis

rleonid:
Good point!

diml:
thanks

PR #198, opened 2017-04-05
Replace handwritten ppf with Format.formatter_of_buffer call

rbonichon:
- Behavior is identical
- Makes utop forward compatible with merged PR 595 of ocaml trunk

Also it is shorter

### DIFF
--- a/src/lib/uTop.ml
+++ b/src/lib/uTop.ml
@@ -145,11 +145,10 @@ let collect_formatters buf pps f =
       pps save
   in
   (* Output functions. *)
-  let out_string str ofs len = Buffer.add_substring buf str ofs len
-  and out_flush = ignore
-  and out_newline () = Buffer.add_char buf '\n'
-  and out_spaces n = for i = 1 to n do Buffer.add_char buf ' ' done in
-  let out_functions = { Format.out_string; out_flush; out_newline; out_spaces } in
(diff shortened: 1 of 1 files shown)


src/lib/uTop.ml
```diff
@@ -145,11 +145,8 @@ let collect_formatters buf pps f =
       pps save
   in
   (* Output functions. *)
-  let out_string str ofs len = Buffer.add_substring buf str ofs len
-  and out_flush = ignore
-  and out_newline () = Buffer.add_char buf '\n'
-  and out_spaces n = for i = 1 to n do Buffer.add_char buf ' ' done in
-  let out_functions = { Format.out_string; out_flush; out_newline; out_spaces } in
+  let ppb = Format.formatter_of_buffer buf in
+  let out_functions = Format.pp_get_formatter_out_functions ppb () in
```
diml: Since `ppb` is only used to get the out functions, I think it would be better to write:

```ocaml
let out_functions =
  let ppb = ... in
  Format.pp...
in
```

The old code wasn't written this way, but we might as well improve it

rbonichon:
I've pushed (force) the requested changes to my branch. (should have checked on GitHub beforehand)

diml:
Merged, thanks!

PR #199, opened 2017-04-08
Stashing updated

copy:
I picked up @chrismamo1's progress and added the remaining bits:

- Stripping of colour escapes
- Separate inputs and outputs in the stashable history
- Another directive besides `#utop_stash`: `#utop_save`, which saves the session as if it was entered in a simplified prompt:

```
# let x = 5;;
val x : int = 5
# let y = 6;;
val y : int = 6
# x + y;;
- : int = 11
# x ^ y;;
Error: This expression has type int but an expression was expected of type
         string
# !@#$%;;
Error: Syntax error
```

Warnings are omitted currently, but I don't consider this a vital feature.

### DIFF
--- a/src/lib/uTop_history.ml
+++ b/src/lib/uTop_history.ml
@@ -1,0 +1,70 @@
+(*
+ * uTop_history.ml
+ * -----------------
+ * Copyright : (c) 2017, Fabian Hemmer <copy@copy.sh>
+ * Licence   : BSD3
+ *
+ * This file is a part of utop.
+ *)
+
+type entry =
+  | Input of string
+  | Output of string
+  | Error of string
+  | Warnings of string
+  | Bad_input of string
+and t = entry list ref
+
+let create () : t =
+  ref []
+
+let contents (t : t) =
+  !t
+
+let strip_colors s =
+  let len = String.length s in
+  let find_escape offset =
+    try
+      let i = String.index_from s offset '\027' in
+      if i = len - 1 || s.[i + 1] <> '[' then
+        None
+      else
+        Some i
+    with
+    | Not_found -> None
+  in
+  let rec find_color_escapes offset =
+    match find_escape offset with
+    | None -> [offset, len]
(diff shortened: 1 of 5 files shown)


src/lib/uTop_history.ml
```diff
@@ -0,0 +1,70 @@
+(*
+ * uTop_history.ml
+ * -----------------
+ * Copyright : (c) 2011, Jeremie Dimino <jeremie@dimino.org>
```
diml: You should put your name here

src/lib/uTop_history.mli
```diff
@@ -0,0 +1,40 @@
+(*
+ * uTop_history.mli
+ * -------
+ * Copyright : (c) 2011, Jeremie Dimino <jeremie@dimino.org>
```
diml: You should put your name here as well

src/lib/uTop_history.ml
```diff
@@ -0,0 +1,70 @@
...
+let add history v =
+  history := v :: !history
+
+let add_input history input =
+  add history @@ Input (String.trim input)
+
+let add_output history output =
+  let output = String.trim output in
+  if output <> "" then (* directives produce empty output *)
+    add history @@ Output output
```
diml: Why don't we strip colors here?

diml:
BTW, I don't think this work in emacs? Could you either add support for it or make the directives print an error when `UTop.get_ui () = Emacs`?

src/lib/uTop_history.ml
copy: Output is added before colours are added.

diml: ok

diml:
The patch looks good to me. Thanks to you and chrismamo1

diml:
BTW, I just noticed that the directive is registered as `"utop_save_session"`, I'm changing it to `"utop_save"`

Issue #161, opened 2016-03-22
Bind unbound values at the toplevel


ihodes:
E.g. 

``` ocaml
> Uri.parse "some uri";;
> Uri.port _1;;
```

diml:
Is that what you are looking for?

```ocaml
let x = Uri.parse "some uri";;
```

ihodes:
No, I'm aware of let-bindings :) 

Other REPLs have this feature (e.g. iPython (`_1`, `_2`, etc.) and Clojure (`*1`, `*2`, etc). It comes in handy a lot in my experience, and saves a good amount of typing in general.

diml:
I see. That's seems like a nice feature indeed

diml:
Implemented by #200

PR #200, opened 2017-04-13
Bind expressions to _0, _1, etc.

copy:
This rewrites toplevel expressions like `x;;` to `let _n = x;;`, for example:

```ocaml
# let x = 5;;
val x : int = 5
# x + 6;;
val _1 : int = 11
# _1 - 6;;
val _2 : int = 5
```

Only tested on 4.04. I think this is the right approach. The newly created bindings are not available in the output of `#utop_stash`, but I think this unavoidable.

Currently the binding (`val _1 : …`) only shows up if `-show-reserved` is passed. We could show `_n` when `n` is numeric even when reserved identifiers should be hidden.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,9 @@
 1.20 (in preparation)
 ---------------------
 
+* Add `-implicit-bindings` option to automatically bind expressions to names
+  `_0`, `_1` and so on. For example, `3 + 4;;` becomes `let _0 = 3 + 4;;`
+  (#161, #193, Fabian Hemmer)
 * Add tab completion for `#mod_use` (#181, Leonid Rozenberg)
 * Mention `#help` in `#utop_help` (#190, Fabian Hemmer)
 * Add `#utop_stash` and `#utop_save` to save the session to a file
--- a/src/lib/uTop.ml
+++ b/src/lib/uTop.ml
@@ -70,6 +70,7 @@ type syntax =
   | Camlp4r
 
 let hide_reserved, get_hide_reserved, set_hide_reserved = make_variable true
+let create_implicits, get_create_implicits, set_create_implicits = make_variable false
 let show_box, get_show_box, set_show_box = make_variable true
 let syntax, get_syntax, set_syntax = make_variable Normal
 let phrase_terminator, get_phrase_terminator, set_phrase_terminator = make_variable ";;"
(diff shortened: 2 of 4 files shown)


diml:
Thanks, this implements #161.

> Currently the binding (val _1 : …) only shows up if -show-reserved is passed. We could show _n when n is numeric even when reserved identifiers should be hidden.

Does this mean that the output is completely suppressed by default?

copy:
> Does this mean that the output is completely suppressed by default?

Yes. So we have to either change the behaviour of `-show-reserved` or pick another binding name. I like `_n` because it's unlikely to collide with other names, but suggestions are welcome.

diml:
Well, I think unless `-show-reserved` is passed, the output should be the same as before:

```ocaml
# let x = 5;;
- : int = 5
# _1;;
- : int = 5
```

And if you want to see what is what, you'd have to either pass `-show-reserved` or use a prompt that shows the command count.

For the implementation, maybe you could rewrite a topleval `<expr>` into: `let _n = <expr>;; _n;;`. Hopefully that should do the trick

copy:
> Well, I think unless -show-reserved is passed, the output should be the same as before:

I thought it was nice that the user is made aware of the fact that the binding was created implicitly. `- : int = 5` seems to imply that no binding was made.

> For the implementation, maybe you could rewrite a topleval <expr> into: let _n = <expr>;; _n;;. Hopefully that should do the trick

Currently the expression isn't printed in this case:

```
# let _n = 5;; _n;;
# let x = 5;; x;;
val x : int = 5
```

diml:
> I thought it was nice that the user is made aware of the fact that the binding was created implicitly. - : int = 5 seems to imply that no binding was made and is not less verbose than _0 : int = 5.

Indeed, however I'm pretty sure that some people will complain that the output is not the same as the ocaml toplevel and that it's confusing. What we can do is have a separate option to show them, independently of the `-show-reserved` setting. If it turns out that it is the right default, then we'll switch the default in a subsequent release

> Currently the expression isn't printed in this case:

Yh, it's an odd behavior of the lexer, it drops everything after the first `;;`. However if you generate a list of 2 structure items it should be fine (I think)

copy:
> Yh, it's an odd behavior of the lexer, it drops everything after the first ;;. However if you generate a list of 2 structure items it should be fine (I think)

I tried and it doesn't print the trailing expression. Should we go with a flag that disables this feature entirely (for people who want "ocaml toplevel" behaviour)?

copy:
@diml Anything missing to get this merged?

diml:
Nope, looks good. Thanks for reminding me, I missed the last commits

PR #204, opened 2017-04-20
Refactor hooks for better handling of phrase eval, closes #201.

### LINKED ISSUE #201: Define meaning of utop's emacs hooks
Hello folks/maintainer,

I have a compatibility layer to `utop` (actually `rtop`) in Reason and I am basically defining:

```
 ;;; utop.el
  (setq utop-discover-phrase 'reason-discover-phrase-debug)
  (setq utop-skip-blank-and-comments 'reason-skip-blank-and-comments)
  (setq utop-skip-to-end-of-phrase 'reason-find-phrase-end)
```

The problem is that I cannot understand clearly the semantic of `utop-skip-to-end-of-phrase`. I see [here](https://github.com/diml/utop/blob/master/src/top/utop.el#L854) that it is called to set a new `end` of the phrase, but if the triple returned by `utop-discover-phrase` is already returning the end position as "point after the semicolon", then why do we need to skip again?
The result of the above in my case basically is that two phrases are evaluated, unless I never skip anything in my `utop-skip-to-end-of-phrase`).

Maybe I am missing something here :smile: 

Thanks for reading and I hope the issue description is understandable enough.

arichiardi:
The patch tries to improve on the current way of evaluating a phrase and
skipping to the next. It removes both `utop-skip-to-end-of-phrase` and
`utop-skip-blank-and-comments`, but adds `utop-next-phrase-beginning`.
We assume `utop-discover-phrase` is always returning the end position as
at end of a `;;`. Thereafter we use `utop-next-phrase-beginning` to skip
to the next phase if `utop-skip-after-eval-phrase` is non-nil.

### DIFF
--- a/src/top/utop.el
+++ b/src/top/utop.el
@@ -217,14 +217,13 @@ Useful as file variable."))
    "Name of preprocesor. Currently supported camlp4o, camlp4r.
 Useful as file variable."))
 
-(defvar utop-skip-blank-and-comments 'utop-compat-skip-blank-and-comments
-  "The function used to skip blanks and comments.")
-
-(defvar utop-skip-to-end-of-phrase 'utop-compat-skip-to-end-of-phrase
-  "The function used to find the end of a phrase")
+(defvar utop-next-phrase-beginning 'utop-compat-next-phrase-beginning
+  "The function used to find the beginning of the next phrase.")
 
 (defvar utop-discover-phrase 'utop-compat-discover-phrase
-  "The function used to discover a phrase")
+  "The function used to discover a phrase.
+It should return a triple (begin-pos, end-pos,
+end-pos-with-comments)." )
 
 (defvar utop-skip-after-eval-phrase t
   "Whether to skip to next phrase after evaluation.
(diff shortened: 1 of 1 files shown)


src/top/utop.el
```diff
@@ -247,19 +246,12 @@ list of 3 function symbols: (tuareg-symbol typerex-symbol caml-symbol)."
...
-(defun utop-compat-skip-blank-and-comments ()
-  (funcall
-   (utop-compat-resolve '(tuareg-skip-blank-and-comments
-                          typerex-skip-blank-and-comments
-                          caml-skip-blank-and-comments))))
-
-(defun utop-compat-skip-to-end-of-phrase ()
+(defun utop-compat-next-phrase-beginning ()
   (funcall
    (utop-compat-resolve '(tuareg-skip-to-end-of-phrase
```
diml: Shouldn't this be wired to `tuareg-next-phrase`?

src/top/utop.el
```diff
@@ -848,10 +840,8 @@ If ADD-TO-HISTORY is t then the input will be added to history."
          (save-excursion
```
diml: This block can be simplified now

src/top/utop.el
arichiardi: Definitely

arichiardi:
Fixed

diml:
Thanks, looks good

PR #206, opened 2017-04-24
Add support for Reason (native)

arichiardi:
Include the necessary hooks for phrase detection in Reason.

We'll need to wait on: https://github.com/facebook/reason/pull/1227

This is the configuration that I use at the moment:

```
(add-hook 'reason-mode-hook
          (lambda ()
            (utop-minor-mode)
            (define-key utop-mode-map (kbd "C-<up>") 'utop-history-goto-prev)
            (define-key utop-mode-map (kbd "C-<down>") 'utop-history-goto-next)
            (setq utop-command "opam config exec -- rtop -emacs")))
```

I was wondering if the `utop-command` should be set by `utop.el`. What do you think?

### DIFF
--- a/src/top/utop.el
+++ b/src/top/utop.el
@@ -242,6 +242,7 @@ list of 3 function symbols: (tuareg-symbol typerex-symbol caml-symbol)."
    ((eq major-mode 'tuareg-mode  ) (nth 0 choices))
    ((eq major-mode 'typerex-mode ) (nth 1 choices))
    ((eq major-mode 'caml-mode    ) (nth 2 choices))
+   ((eq major-mode 'reason-mode  ) (nth 3 choices))
    (t (error (format "utop doesn't support the major mode \"%s\". It
 supports caml, tuareg and typerex modes by default. For other
 modes you need to set these variables:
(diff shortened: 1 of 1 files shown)


arichiardi:
I need to add a couple of lines to the Readme as well.

arichiardi:
Newbie question, say I have ocaml libs installed somewhere else than `~/.opam`, e.g.; node_modules`. How can I tell utop to consider those as well?

diml:
There are several ways to do that, the easiest is to set the environment variable `OCAMLPATH`. This variable and other related ones are described in the man page for findlib.conf

arichiardi:
Thanks will try that!

arichiardi:
Sorry if this discussion is digressing, but now I understand the moving parts a bit better.
For Reason and Javascript, there is a problem. Even if `findlib` can find the library, the `.ml` or `.re` files that BuckleScript requires use custom things. They need to be compiled to Javascript and cannot be compiled to native code.

So the idea would be to hijack `utop`'s `parse_toplevel_phrase` (the Read phase in a repl) so that it can transpile to Javascript using the BuckleScript compiler, and then the `utop`'s `eval` so that it can actually shell out and execute in a [node repl](https://nodejs.org/api/repl.html). The repl would hold the state and do the side effects.

Not sure how easy all this is and it is purely immaginative, any consideration is very welcome :D

diml:
That seems complicated. Though I have no objection if you need to add some hooks to utop to be able to do that

arichiardi:
Uhm, I checked the `utop` source and you're basically using comint. I like it, and this means that a Reason top for node.js would "just" need to understand the `utop` protocol. Am I right?

diml:
Yh, when I wrote the utop mode I didn't like the fact that with comint you could erase the prompt or edit the output of the toplevel so I developed my own mode which behaved more like a shell. However you are right that implementing the utop protocol would be enough to make it work with the emacs mode.

arichiardi:
Just wanted to link here to my (approved) split of [reason-mode](https://github.com/arichiardi/reason-mode) which fully supports this PR.

arichiardi:
@diml what do you think, shall we merge this one or you prefer to wait a bit?

diml:
Happy to merge now

PR #210, opened 2017-05-30
Fix installation to handle utop.el too

AltGr:
seems that was forgotten in the migration to jbuild. Unfortunately, this causes 'opam-user-setup' to break badly (at emacs startup, it finds the opam package, then breaks trying to load the file, and consequently doesn't load remaining modules)

### DIFF
--- a/jbuild
+++ b/jbuild
@@ -4,6 +4,10 @@
  ((section share)
   (files (utoprc-dark utoprc-light))))
 
+(install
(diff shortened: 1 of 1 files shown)


diml:
Thanks, I will do a bugfix release

diml:
https://github.com/ocaml/opam-repository/pull/9350

PR #215, opened 2017-08-04
Add camomile dep to utop

rgrinberg:
Right now it's being pulled transitively from zed, but it's being used directly
in utop


diml:
Thanks

PR #216, opened 2017-08-18
Some random work I did on the emacs mode

rgrinberg:
I got rid of all the old Emacs workarounds. Emacs 24 is 5 years old, it's a
reasonable minimum. Also cleaned up a bunch of code that is more tersely done
with ocaml like pattern matching (pcase, which is available for emacs >= 24).

Finally, I've hacked together a simple company backend for utop completion. It's
not really ready for merging, but if there's interest I can work on it further.
For example: make the old completion still available for those who don't
use/like company, add type signatures to completions.

I've found it easier to implement the completion by adding a simple completion
command in utop itself. Although that might because I don't quite understand all
the details in the old command. Would appreciate some comments here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 _build/
 .merlin
 *.install
+*.elc
--- a/src/top/utop.el
+++ b/src/top/utop.el
@@ -244,7 +244,7 @@ list of 3 function symbols: (tuareg-symbol typerex-symbol caml-symbol)."
    ((eq major-mode 'caml-mode    ) (nth 2 choices))
    ((eq major-mode 'reason-mode  ) (nth 3 choices))
    (t (error (format "utop doesn't support the major mode \"%s\". It
-supports caml, tuareg and typerex modes by default. For other
+supports caml, tuareg, typerex and reason modes by default. For other
 modes you need to set these variables:
 
 - `utop-next-phrase-beginning'
--- a/src/top/utop.el
+++ b/src/top/utop.el
@@ -400,7 +400,7 @@ it is started."
      ('copy "You cannot edit the buffer while waiting for copy of last input")
      ('hist "You cannot edit the buffer while waiting for history"))))
 
-(defun utop-before-change (start stop)
+(defun utop-before-change (_start stop)
   (unless utop-inhibit-check
     (cond
      ((not (eq utop-state 'edit))
@@ -626,7 +626,7 @@ it is started."
          (minibuffer-hide-completions))
        (setq utop-completion nil)))))
 
-(defun utop-process-output (process output)
+(defun utop-process-output (_process output)
   "Process the output of utop"
   (with-current-buffer utop-buffer-name
     (utop-perform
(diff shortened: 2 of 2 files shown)


tuncer:
What's the progress with this branch?

diml:
TBH I don't have much time to spend on utop these days so I haven't reviewed this patch

rgrinberg:
I'm using this patch and it works well for me. I think we can safely merge all up to the last commit in this branch without any issues.

For the last commit, I think it's just a matter of whether we want to maintain the old completion mechanism. If @diml wants to keep it, I can refactor the code so that both options are available. It would also be nice to support something other than company, but AFAIK, nothing else support async generation of completion candidates.

diml:
> For the last commit, I think it's just a matter of whether we want to maintain the old completion mechanism.

I don't mind getting rid of it if the new way is better.

Basically, I don't mind merging this PR, but I ask that someone joins the project and accept to take care of any follow up work.

rgrinberg:
OK, so let me cherry pick everything but the first commit. I'll re-work the last commit to make the company backend optional. It is vastly better, but it requires a 3rd party dependency (company-mode). I'll have another look at defaulting to defaulting to pcomplete or something else that is less intrusive.

rgrinberg:
Oh, and I'm definitely willing to help maintain the Emacs mode.

Btw, can you at least take a look at the company specific utop command that I've added. Just to make sure I'm not missing something obvious. I couldn't get the completion to work with the original code.

diml:
Thanks, BTW, personally I don't mind if it doesn't work without company-mode, especially if supporting only one mode makes the code simpler.

The `complete-company` command implementation looks fine to me.

Feel free to merge at your convenience :)

rgrinberg:
OK, thanks. I will take another look at simplifying the completion stuff. Another issue I thought is that a lot of people get their elisp from melpa. Which means that their version of utop.el might not be compatible with the version of utop they have installed. I'll try and at least handle that more gracefully, even if we have 1 completion mode.

rgrinberg:
I moved the completion stuff to #233

diml:
Maybe the emacs mode and utop could exchange the version of the protocol they implement at the beginning of the session?

rgrinberg:
Yeah, that would be ideal. We still need some code to handle the currently released utop though. As it's not going to have any knowledge of this.

diml:
Indeed

tuncer:
@diml wrote:

> TBH I don't have much time to spend on utop these days so I haven't reviewed this patch

From what I gather, utop is quite popular and one might call it the toplevel of choice for most developers. Thus, have you considered asking for extra collaborators or thought about moving it to an org like https://github.com/ocaml, https://github.com/ocamllabs?

diml:
Well, I asked about a new maintainer on the caml list but actually maybe it's a too big project to be just handed over and I should look for more collaborators first.

About moving it somewhere else, I vaguely remember we talked about it at some point but I can't remember why it didn't happen. /cc @gasche @avsm, any objection to moving utop to github.com/ocaml?

gasche:
No objections from me; I think there has been recently a push to sub-organisational-repos, so something like ocaml-tools could be proposed and would be equally appropriate (Merlin could go there as well), but as far as I'm concerned ocaml/ is fine.

diml:
Ok, well I guess we can move it to ocaml/ for now and see later if we want to move things somewhere else

PR #233, opened 2018-02-26
Company mode

rgrinberg:
@tuncer if you use company mode, would you mind giving this branch a try?

### DIFF
--- a/src/top/utop.el
+++ b/src/top/utop.el
@@ -234,8 +234,11 @@ Caml toplevel")
   "continuation function to populate the candidates for the company
 backend")
 
-(defvar utop-version "0"
-  "detected version of utop. 0 for unknown")
+(defvar utop-protocol-version "0"
+  "detected version of utop protocol. 0 for unknown or version pre")
+
+(defvar utop--read-version nil
+  "whether we've tried to detect the utop version")
 
 (defvar utop--company-loaded nil)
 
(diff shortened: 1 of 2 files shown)


rgrinberg:
OK I've made the company support completely optional to unbreak older utops.

tuncer:
I hadn't configure utop.el yet. I can give it a try if you're still in need of basic testing. I don't think I would test advanced features if I did.

rgrinberg:
I just thought that you wanted to give this feature a try since you asked about the original PR. I've tested this feature myself well enough so it's not necessary for you to do it as well unless you're comfortable with it.

rgrinberg:
@diml I'm parsing `utop -version` to detect which utop we're dealing with. Not a thing of beauty, but it will do for now.

diml:
What about saying that the first line utop prints when started with `utop -emacs` is `protocol-version:N`, and if it doesn't print it then the emacs mode assumes that it's using protocol version 1?

rgrinberg:
OK sure, that's done.

src/top/utop.el
```diff
@@ -635,10 +656,18 @@ it is started."
...
-         ;; Remove it and continue
-         (setq lines (cdr lines)))
+         ;; process first line. if we haven't tried reading the version, we'll
+         ;; trying to do it now.
+         (let ((line (car lines)))
+           (if utop--read-version
+               (utop-process-line line)
+             (progn
+               (save-match-data
+                 (when (string-match "protocol-version:\\([0-9]+\\)" line)
```
diml: If the line doesn't match, do we still process it normally or does it end up being discarded?

rgrinberg: Good point. I'll fix that.

rgrinberg:
Will release 2.1.0 with this change.

PR #236, opened 2018-06-11
Company backend only complete on utop buffers

marsam:
Change the company backend to only work on utop buffers, otherwise it does not passes control to the next backend

### DIFF
--- a/src/top/utop.el
+++ b/src/top/utop.el
@@ -1221,7 +1221,8 @@ Special keys for utop:
   (pcase command
     ('interactive (company-begin-backend 'utop-company-backend))
     ('sorted t)
-    ('prefix (company-grab-symbol))
+    ('prefix (and (derived-mode-p 'utop-mode)
+                  (or (company-grab-symbol-cons "\\." 1) 'stop)))
     ('candidates
      (progn
        `(:async


pmetzger:
Sorry for not looking at this earlier. I'm afraid I don't understand what it does. Can you explain it more?

rgrinberg:
I reviewed this change earlier and it looks like my comment disappeared somehow. Sorry about that.

My main question about this change is: why is the check only applied to the `'prefix` command? Seems like it would make sense to apply to all commands. @Wilfred do you know perhaps?

Wilfred:
There's not much company documentation (I'm only aware of [this example backend](https://github.com/company-mode/company-mode/wiki/Writing-backends) and [this blog post](http://sixty-north.com/blog/writing-the-simplest-emacs-company-mode-backend)), but that blog post states:

> The prefix command is probably the most complex command to handle. This command should return the text that is to be completed. Determining this text can be complex depending on what you're trying to complete, but company-grab-symbol often does "the right thing" if your completion context is space-delimited.
>
> If the prefix command returns nil, this tells company-mode that the backend is not suitable for doing completion on this context.

So I think this is the correct approach.

Wilfred:
Aha, there's a docstring on `company-backends` which says:

> prefix: The backend should return the text to be completed.  It must be
text immediately before point.  Returning nil from this command passes
control to the next backend.

pmetzger:
Merged.

PR #238, opened 2018-06-20
Minimal update to make utop build on OCaml 4.07.0+rc1

hcarty:
This is not the _right_ fix.  It just allows `make` to complete.

### DIFF
--- a/src/lib/uTop_complete.ml
+++ b/src/lib/uTop_complete.ml
@@ -541,6 +541,25 @@ let list_global_names () =
     | Env.Env_constraints (summary, _) ->
         loop acc summary
 #endif
+#if OCAML_VERSION >= (4, 07, 0)
+    | Env.Env_copy_types (summary, _) ->
+        loop acc summary
+#endif
+#if OCAML_VERSION >= (4, 07, 0)
+    | Env.Env_open(summary, _skip, path) ->
+        match try Some (Path_map.find path !local_fields_by_path) with Not_found -> None with
+          | Some fields ->
+              loop (String_set.union acc fields) summary
+          | None ->
+              match lookup_env find_module path !Toploop.toplevel_env with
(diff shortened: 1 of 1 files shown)


src/lib/uTop_complete.ml
```diff
@@ -541,6 +541,25 @@ let list_global_names () =
     | Env.Env_constraints (summary, _) ->
         loop acc summary
 #endif
+#if OCAML_VERSION >= (4, 07, 0)
+    | Env.Env_copy_types (summary, _) ->
+        loop acc summary
+#endif
+#if OCAML_VERSION >= (4, 07, 0)
```
diml: The conditions in these two blocks is the same, shouldn't they be conflated?

diml:
Why is not the right fix BTW?

hcarty:
@diml I didn't do much looking into the new/changed constructors so I'm not sure if there are further changes to `uTop_complete.ml` required to handle the constructors properly.  I'm not familiar with utop's internal or [this comment](https://github.com/ocaml/ocaml/blob/4197e75b6b864c787cdf6162c1303dd74e1e3d5b/typing/env.mli#L33-L34) upstream so I don't want to presume that this minimal update is correct. 

I'll simplify the `#if`ing as you suggested [here](https://github.com/diml/utop/pull/238#discussion_r197054574).  If you're happy with the PR at that point I'm all for merging so utop can be ready for 4.07.0!

hcarty:
@diml Simplified - the overall diff is fairly minimal now.

diml:
> I'm not familiar with utop's internal or this comment upstream so I don't want to presume that this minimal update is correct.

Ok. This comment was actually from one of my PR :) I checked utop's code and convinced myself that there is nothing more to do here.

Thanks!

PR #241, opened 2018-07-06
Fix utop-eval-phrase on recent tuareg-mode

Wilfred:
tuareg-next-phrase was moved to tuareg_indent.el in
ocaml/tuareg@b40e733c1 and removed entirely in
ocaml/tuareg@b2ad75f3d5 (September 2017).

As a result, utop-eval-phrase crashes due to the missing function.

Instead, write a small function to move to the next phrase, similar to
the movement logic in tuareg-eval-phrase.

### DIFF
--- a/src/top/utop.el
+++ b/src/top/utop.el
@@ -272,9 +272,19 @@ modes you need to set these variables:
 " major-mode))))
    choices))
 
+(defun utop-tuareg-next-phrase ()
+  "Move to the next phrase after point."
+  (let* ((pos (tuareg--after-double-colon))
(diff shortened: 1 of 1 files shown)


rgrinberg:
Thanks. cc @Chris00 here. I realize, that it was hard to predict this breakage, but it's good to be aware of things when they break.

rgrinberg:
Perhaps we should also stop relying on the private functions from tuareg as well.

Wilfred:
Yep, that thought had crossed my mind. :)

I'm not sure what a good alternative would be, but this gets it working in the meantime.

PR #243, opened 2018-07-07
Clflags.no_std_include will get renamed Clflags.no_stdlib

sliquister:
When https://github.com/ocaml/ocaml/pull/1569 is merged. This should probably not be merged until the ocaml pull request itself is merged.


avsm:
It looks like the upstream OCaml issue got closed...

PR #251, opened 2018-08-23
Make more functions tail-recursive, and use tail-rec List functions

gpetiot:
The majority of changes are straightforward, but the commit 826915d could use a review (not sure if it really improves the situation and we lost in readability).


src/lib/uTop.ml
```diff
@@ -104,7 +105,8 @@ let default_keywords = [
   "when"; "while"; "with"; "try_lwt"; "finally"; "for_lwt"; "lwt";
 ]
 
-let keywords = ref (List.fold_right String_set.add default_keywords String_set.empty)
+let add set elt = String_set.add elt set
+let keywords = ref (List.fold_left add String_set.empty default_keywords)
```
diml: What about using `String_set.of_list`?

src/lib/uTop_history.ml
```diff
@@ -33,15 +35,18 @@ let strip_colors s =
...
+      match find_escape offset with
+      | None -> (offset, len) :: acc
+      | Some esc_offset ->
+         try
+           let i = String.index_from s esc_offset 'm' in
+           aux ((offset, esc_offset) :: acc) (i + 1)
+         with
+         | Not_found -> (offset, len) :: acc
+    in
+    List.rev (aux [] offset)
```
diml: Note that we do a `List.map` on the result, so we might as well not do this `List.rev` and use `List.rev_map`

src/lib/uTop_styles.ml
```diff
@@ -98,87 +98,107 @@ let load () =
         Lwt_log.error_f "cannot load styles from %S: %s: %s" fn func (Unix.error_message error)
     | exn -> Lwt.fail exn)
 
-let rec stylise_filter_layout stylise tokens =
-  match tokens with
+let stylise_filter_layout stylise tokens =
```
diml: We can probably make this code clearer by using something like `List.filter_map`

src/top/expunge/jbuild
```diff
@@ -1,5 +1,7 @@
 (jbuild_version 1)
 
+(rule (copy ../../lib/uTop_utils.ml uTop_utils.ml))
```
diml: Why not simply use the utop library in expunge?

diml:
I'm not sure about 826915d either: the input of the `stylise` functions comes from the user, so I think we'd overflow the terminal buffer way before we blow up utop's stack. In the end it's better to keep the code simple.

src/lib/uTop_utils.ml
```diff
@@ -0,0 +1,33 @@
...
+  let map2 f l l' =
+    List.rev (List.rev_map2 f l l')
+
+  let concat l =
+    let rec aux acc = function
+    | [] -> acc
+    | h :: t ->
+       aux (List.rev_append h acc) t
+    in
+    List.rev (aux [] l)
```
diml: I'm wondering if we shouldn't just use a stdlib more systematically for community projects. That would make maintaining them easier overall

diml:
That's fine, however we'd have to drop compatibility with OCaml < 4.04

rgrinberg:
That's acceptable. Utop isn't an essential project like dune that keeping strict compatibility with ancient versions of OCaml is necessary.

pmetzger:
If we adopt base, the possibility of ultimately merging this into the main OCaml release goes away, so that might also be a consideration.

pmetzger:
(It would be nice to replace the default toplevel with utop, though I know that's a bit of a dream.)

rgrinberg:
Yeah, I don't think it's desirable to intertwine utop's development with the rest of OCaml anyways.

diml:
So far, we have been trying to take things out of the compiler, so it seems unlikely that utop would go in. There is some value in trying to keep the dependencies of utop to a minimum: it's easier to get it up and running with development compilers. However, there is a maintenance cost as well.

In general, using a stdlib for community maintained projects seems like a win to me as it avoids maintaining the same utility functions in many places.

pmetzger:
> In general, using a stdlib for community maintained projects seems like a win to me as it avoids maintaining the same utility functions in many places.

Sure, but what's missing from the default stdlib that utop needs at this point?

diml:
This PR adds a `UTop_utils` module that looks like a mini-stdlib. Looking at it again, it simply redefines a couple of functions from the stdlib to be tail-rec, so it doesn't seem worth having it in utop. 

@gpetiot could you remove `UTop_utils` from this PR? There is nothing in utop that would justify not using the standard versions.

src/lib/uTop_styles.ml
```diff
@@ -98,21 +98,20 @@ let load () =
         Lwt_log.error_f "cannot load styles from %S: %s: %s" fn func (Unix.error_message error)
     | exn -> Lwt.fail exn)
 
-let rec stylise_filter_layout stylise tokens =
-  match tokens with
-    | [] ->
-        []
-    | (Comment (Comment_reg, _), loc) :: tokens ->
+let stylise_filter_layout stylise tokens =
+  let rec aux acc = function
```
diml: This function doesn't look recursive anymore

diml:
Merged, thanks!

PR #265, opened 2018-11-23
Add support for -dtypedtree

Armael:


Armael:
Well, this made me realize that the types are not printed, which makes the flag pretty much useless...

diml:
If you look at where `dump_parsetree` is used in the code of utop, you can probably add some code to interpret `dump_typedtree` as well.

PR #269, opened 2019-01-15
Automatically install printers marked with [@@ocaml.toplevel_printer]

diml:
This PR makes utop automatically install as printer all functions marked with `[@@ocaml.toplevel_printer]`  in mli files. This gives a very simple way for users to define new toplevel printers.


avsm:
This is a good idea! Still failing on 4.03 with

```
- File "src/lib/uTop_main.ml", line 615, characters 22-51:
- Error: Unbound module Env.Persistent_signature
```

diml:
Oops, indeed I got the condition wrong. It's fixed now.

pmetzger:
Looks like all checks are successful, so I'll merge this.

PR #271, opened 2019-01-31
Prepare release 2.3.0

avsm:


avsm:
The build failures on macOS are due to an unrelated failure due to the opam 2.0.3 release. Manually tested.

PR #278, opened 2019-04-27
unicode support improvements: wider glyph, combined glyph

kandu:
note, this pr depends on another pr in the lambda-term project ocaml-community/lambda-term#71

* UTop_lexer: support combined glyph
* follow the api changes of zed_char and zed_string

### DIFF
--- a/src/lib/uTop.ml
+++ b/src/lib/uTop.ml
@@ -441,7 +441,7 @@ let make_prompt ui profile count size key_sequence (recording, macro_count, macr
               txta;
               Array.make
                 (size.cols - Array.length txta - Array.length txtb)
-                (UChar.of_int 0x2500, { none with foreground = Some (color lcyan blue); bold = Some bold });
+                (Zed_char.unsafe_of_uChar (UChar.of_int 0x2500), { none with foreground = Some (color lcyan blue); bold = Some bold });
               txtb;
             ]
         ) second_line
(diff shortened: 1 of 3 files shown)


pmetzger:
I'm not sure how good I would be at reasonably reviewing this. This brings up the question of who should, or whether we need review before committing.

diml:
I'd say it depends of the quality of service we want to provide for utop. In my opinion, utop is not a fundamental building block of the ecosystem: it is a tool human use to play with libraries but it is not something they rely on for building the software they put in production.

The most important thing to care about here is the user experience: if the behaviour of utop keeps changing all the time then that's annoying for users. Apart from that, I personally feel relax about letting motivated contributors develop their feature as long as they are reactive about providing follow up support.

However, one thing to keep in mind is the recent npm security debacle. I.e., when new contributors show up, we should eyeball their patches until we trust them.

kandu:
For these particular PRs, we can pay attention only on the API changes. Implementation bugs are easy to fix, but a design mistake in the API will annoy us for a long period of time until next major release.

pmetzger:
> if no one else feels comfortable reviewing such patches, I feel like we need to trust a bit more contributors who look motivated by the project

I don't disagree. I just don't want to make that decision entirely on my own. :)

pmetzger:
> if no one else feels comfortable reviewing such patches, I feel like we need to trust a bit more contributors who look motivated by the project

I don't disagree, but I don't think I should be making that decision on my own.

diml:
That's a fair point. Though, we need to find a good balance. I'm personally finding it hard to find time for utop, and if no one else feels comfortable reviewing such patches, I feel like we need to trust a bit more contributors who look motivated by the project. It's a small risk anyway, if it doesn't work out we can always revert the patches.

diml:
I don't disagree either. I feel like for a project like utop, it is fine to have a more dynamic approach and rely on publishing bugfixes quickly.

@kandu, I suggest that you merge this patch once you are satisfied with your testing. I had a quick look and didn't see anything chocking.

pmetzger:
So if this is merged now, should we consider a new release?

kandu:
I think It's time to make a new release.

pmetzger:
So please do!

PR #282, opened 2019-05-17
utop release 2.4.0

kandu:


kandu:
@diml, @pmetzger any suggestions about there PRs for the new releases?
If everything is fine, would you mine having a glance at them and approve.

CHANGES.md
```diff
@@ -1,3 +1,10 @@
+2.4.0 (2019-05-17)
```
diml:
```suggestion
2.4.0 (2019-05-17)
------------------
```

(to follow the style of older releases)

CHANGES.md
```diff
@@ -1,3 +1,10 @@
...
+
+* UTop: depend on lambda-term 2 and therefore provide better unicode support(wide, combined glyph).
+* UTop\_lexer: support combined glyph
```
diml: I would suggest to only mention the parts that are public in the changelog. `UTop_lexer` is an internal module. The role of the changelog is to give users an overview of changes that might affect them.

So for instance, I would summarize the two above points into:

```
* Better support for wide and combined glyph (<your name or github id>, #NNN)
```

I just made up the sentence BTW, I'm not a unicode expert so I might have got it wrong. You can mention `UTop` if its API changed since it is a public module.

diml:
Just a couple of suggestions regarding the changelog, but otherwise it looks good. I had a quick look at zed and lambda-term and the releases look good.

BTW, I noticed that utop is not compatible with 4.08, do you mind looking if it can be easily fixed? In general, it's better if new releases don't have a upper bound on the version of OCaml. Otherwise, users have to choose between using the new utop or the new OCaml.

XVilka:
Agree, it makes sense to wait a bit with a release, since 4.08 is almost ready: https://github.com/ocaml/ocaml/milestone/15

kandu:
Thanks a lot for your suggestions, based on which we'll update the CHANGES.md and rearrange for the release schedule.

But before that, it seems we are encountering a serious choise whether to drop support for ocaml < 4.08 since ocaml 4.08 deprecates a large part of the Format module. The interfaces of lambda-term are also deprecated with 4.08, mainly in LTerm_text
* val make_formatter
* val pp_with_style
* val styprintf
* val kstyprintf

diml:
Regarding dropping support for OCaml < 4.08 in a library X at version V, this is the main point one should reason about: how hard is to write code that depends on X and is compatible with `X<V` and `X>=V`. If it's easy, then dropping support for OCaml < 4.08 is not a big deal. If it's not possible because of breaking changes, then users of X will kind of be forced to make the switch and depend on OCaml >=4 .08, otherwise they won't be able to use the new version of X. This is not necessarily a bad thing, but sometimes people are stuck with an older OCaml for reasons outside of their control.

So here the important question is: do we have breaking changes in the `UTop` module? The `UTop` module is public and people might use in their `~/.ocamlinit` files. Regarding lambda-term and zed, the same question arise, however these two libraries have much less users in opam, so the damage is much smaller.

kandu:
The breaking changes will only occur in lambda-term. But in my opinion, it's really a pity to make lambda-term 2 available only with ocaml >= 4.08. I'd rather use these deprecated function in the Format module and keep lambda-term available with all ocaml versions thus people working with old ocaml will gain the benefit of better unicode support from the upcoming utop 2.4 and lambda-term 2.

p.s. now zed and lambda-term are compatible with 4.08, commits will be pushed later.

kandu:
and to make utop compatible with 4.08 https://github.com/ocaml-community/utop/pull/284

kandu:
lambda-term will still be compatible with ocaml >= 4.02, depending on some deprecated functions since 4.08.

Since utop is now compatible with 4.08, and all change requests are applied, if there is no objection, I'll merge all three PRs in 36 hours and release them in the official opam repo.

PR #284, opened 2019-05-23
compatible with 4.08

kandu:
this is a quick fix to make utop compatible with ocaml 4.08
since I'm not quite familiar with the internal compiler, this PR can serve as a trivial clue for further works.

### DIFF
--- a/src/lib/uTop.ml
+++ b/src/lib/uTop.ml
@@ -803,13 +803,13 @@ let () =
 let () =
   (* "utop" is an internal library so it is not passed as "-package"
      to "ocamlfind ocamlmktop". *)
-  try Topfind.don't_load_deeply ["utop"]; with Fl_package_base.No_such_package _-> ();
+  Topfind.don't_load_deeply ["utop"];
   Topfind.add_predicates ["byte"; "toploop"];
   (* Add findlib path so Topfind is available and it won't be
      initialized twice if the user does [#use "topfind"]. *)
   Topdirs.dir_directory (Findlib.package_directory "findlib");
   (* Make UTop accessible. *)
-  try Topdirs.dir_directory (Findlib.package_directory "utop") with Fl_package_base.No_such_package _-> ()
+  Topdirs.dir_directory (Findlib.package_directory "utop")
 
 (* +-----------------------------------------------------------------+
    | Deprecated                                                      |
--- a/src/lib/uTop.ml
+++ b/src/lib/uTop.ml
@@ -811,6 +811,18 @@ let () =
   (* Make UTop accessible. *)
   Topdirs.dir_directory (Findlib.package_directory "utop")
 
+ (* +-----------------------------------------------------------------+
+   | Compiler-libs re-exports                                        |
+   +-----------------------------------------------------------------+ *)
+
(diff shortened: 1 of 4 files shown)


kandu:
the only breaking change is the elimination of re-exporting load_path= Config.load_path.

In my opinion, the easiest way to get it back and make it compatible with all ocaml versions is to add two functions: get_load_path & set_load_path, then for 4.08 we map the two functions to the new 4.08 compiler-libs api and for ocaml < 4.08, we just have to get/set the value of (Config_load_path: string list ref) within the two new functions.

diml:
I had a quick look and the patch looks correct to me. However, the `try ... with Fl_package_base.No_such_package _-> ()` blocks look suspicious.

> In my opinion, the easiest way to get it back and make it compatible with all ocaml versions is to add two functions: get_load_path & set_load_path, then for 4.08 we map the two functions to the new 4.08 compiler-libs api and for ocaml < 4.08, we just have to get/set the value of (Config_load_path: string list ref) within the two new functions.

Indeed. Ideally we would put all the assumptions we make about the compiler libraries in a library between utop and the compiler. More precisely:

- we would build utop with `(implicit_transitive_deps false)` in the `dune-project` file
- we would introduce a new library `utop_compiler_assumptions` such that:
  + `utop_compiler_assumptions` is the only library in the repository that is allowed to depend directly on any `compiler-libs.xxx` library
  + all other libraries are only allowed to depend on `utop_compiler_assumptions`

If we did this, then this library would be the only thing that would dispatch on the compiler version and would be the only library to update when the compiler libraries change. It would basically serve as a connector between utop and the compiler. It could also serve as a basis to improve the toplevel libraries in the compiler.

kandu:
> However, the `try ... with Fl_package_base.No_such_package _-> ()` blocks look suspicious.

This is for testing convenience, utop is not installable before this patch, it can be safely removed.

What and how many assumptions will be abstracted into utop_compiler_assumptions? Entire parsetree or just some access point, like load_paths?

diml:
> This is for testing convenience, utop is not installable before this patch, it can be safely removed.

Ok

> What and how many assumptions will be abstracted into utop_compiler_assumptions? Entire parsetree or just some access point, like load_paths?

Whatever we'd need to build in such a setup. The parsetree and typedtree would indeed not be visible in the rest of utop, unless they are re-exported by `utop_compiler_assumptions`. Technically, we probably wouldn't want to re-export all these types. However, a subset of these types could be exposed as abstract types with functions to analyse them or construct them. There should be just enough for what we need in utop.

diml:
You could basically think of `utop_compiler_assumptions` as a shield: the compiler libraries are unstable and change all the time. The role of `utop_compiler_assumptions` would be to add a stable layer on top of these. If you keep the scope of this layer small, i.e. for utop only, it should be manageable.

Then, moving forward when we'd be happy with some API in `utop_compiler_assumptions`, we'd try to move it to the compiler proper, so that the compiler libraries themselves improve and get more stable over time.

diml:
Current patch looks good BTW

kandu:
Thanks. Since this PR is approved, it will be closed and rebased to the release-2.4 branch.

PR #292, opened 2019-07-12
Remove now-broken elisp that depends on utop-ocaml-preprocessor

m-plamann:
A recent commit deleted the variable utop-ocaml-preprocessor in utop.el, but there are still some references to that variable, causing the error
`Symbol’s value as variable is void: utop-ocaml-preprocessor`

This pull request deletes the remaining references to utop-ocaml-preprocessor.

### DIFF
--- a/src/top/utop.el
+++ b/src/top/utop.el
@@ -1023,10 +1023,6 @@ defaults to 0."
 
 (defun utop-hack-local-variables ()
   "Perform actions defined by local variables"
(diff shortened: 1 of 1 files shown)


pmetzger:
Makes sense to me. @diml this is tiny, should I commit?

diml:
Yes, that makes sense to me too

pmetzger:
Merged! Thanks!

PR #299, opened 2019-09-16
Fixes for 4.09

### LINKED ISSUE #298: doesnt compile on 4.09.0+beta2
There are some breaking changes in 4.09 that break compilation of utop:

- Env.Persistent_signature needs to be renamed to Persistent_env.Persistent_signature (just a search-and-replace fixes this
- Obj.truncate is now deprecated, and there is some gnarly looking code in the main loop that blits over Sys.argv using it:
```
  let len = Array.length Sys.argv - !Arg.current in
  Array.blit Sys.argv !Arg.current Sys.argv 0 len;
  Obj.truncate (Obj.repr Sys.argv) len;
```
- [ ] TypedtreeIter.DefaultIteratorArgument needs to be ported to use Tast_iterator instead. I've not looked into what this involves yet -- help welcome

Octachron:
This PR fixes the compilation with OCaml 4.09:

  - The use of `Obj.truncated` for overriding `Sys.argv` is replaced by a call to `caml_sys_modify_argv`
  - Iteration on the typed tree is implemented with `Tast_iterator` rather than `TypedtreeIter`.
  - Loading persistent signature is done with `Persistent_env.Persistent_signature`

Closes #298

### DIFF
--- a/src/lib/uTop_main.ml
+++ b/src/lib/uTop_main.ml
@@ -1284,14 +1284,28 @@ let prepare () =
       Format.eprintf "Uncaught exception: %s\n" (Printexc.to_string exn);
       false
 
+#if OCAML_VERSION >= (4, 09, 0)
+external caml_sys_modify_argv : string array -> unit =
+  "caml_sys_modify_argv"
+let override_argv () =
+  let len = Array.length Sys.argv - !Arg.current in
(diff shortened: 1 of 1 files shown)


pmetzger:
@Octachron, if the CI passes, and you feel you've tested this sufficiently (and it wouldn't take that much I think), feel free to commit it.

PR #309, opened 2020-01-05
Fix color highlight.

chripell:
The output from the compiler changed and now mentions the Line.
Tested in emacs and from the CLI.

### DIFF
--- a/src/lib/uTop.ml
+++ b/src/lib/uTop.ml
@@ -126,9 +126,21 @@ let get_ocaml_error_message exn =
     Scanf.sscanf
       str
       "Characters %d-%d:\n%[\000-\255]"
-      (fun start stop msg -> ((start, stop), msg))
+      (fun start stop msg -> ((start, stop), msg, None))
   with _ ->
-    ((0, 0), str)
+  try
+    Scanf.sscanf
+      str
+      "Line %d, characters %d-%d:\n%[\000-\255]"
+      (fun line start stop msg -> ((start, stop), msg, Some(line, line)))
+  with _ ->
+  try
+    Scanf.sscanf
+      str
+      "Lines %d-%d, characters %d-%d:\n%[\000-\255]"
+      (fun start_line stop_line start stop msg -> ((start, stop), msg, Some(start_line,stop_line)))
+  with _ ->
+    ((0, 0), str, None)
 
 let collect_formatters buf pps f =
   (* First flush all formatters. *)
@@ -196,6 +208,7 @@ let discard_formatters pps f =
    +-----------------------------------------------------------------+ *)
 
 type location = int * int
+type lines = int * int
 
 type 'a result =
   | Value of 'a
@@ -365,10 +378,10 @@ let check_phrase phrase =
           None
         with exn ->
           (* The phrase contains errors. *)
-          let loc, msg = get_ocaml_error_message exn in
+          let loc, msg, line = get_ocaml_error_message exn in
           Toploop.toplevel_env := env;
           Btype.backtrack snap;
-          Some ([loc], msg)
+          Some ([loc], msg, [line])
 
 (* +-----------------------------------------------------------------+
    | Prompt                                                          |
--- a/src/lib/uTop_main.ml
+++ b/src/lib/uTop_main.ml
@@ -79,6 +79,39 @@ let index_of_offset src ofs =
 
 let convert_locs str locs = List.map (fun (a, b) -> (index_of_offset str a, index_of_offset str b)) locs
 
+let convert_loc str (a, b) = (index_of_offset str a, index_of_offset str b)
+
+let get_line src line =
+  let rec aux line' ofs skipped =
+    if ofs >= String.length src then
+      ("", 0)
+    else if line' = line then
+      (String.sub src ofs (String.length src - ofs), skipped)
+    else
+      let ch, next_ofs = Zed_utf8.unsafe_extract_next src ofs in
+      if Zed_utf8.escaped_char ch = "\\n" then
+        aux (line' + 1) next_ofs (skipped + 1)
+      else
+        aux line' next_ofs (skipped + 1)
+  in
+  aux 1 0 0
+
+let convert_one_line str line ofs=
+  let selected, skipped = get_line str line in
+  index_of_offset selected ofs + skipped
+
+let convert_line str (start_ofs, end_ofs) (start_line, end_line) =
+  (convert_one_line str start_line start_ofs,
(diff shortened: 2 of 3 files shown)


pmetzger:
I unfortunately don't have time to look through this myself. Quick question, though: is the change back compatible with older compilers?

chripell:
Yes, I've tested it with 4.05.0 as well, which used to work (has only the "columns" output instead of the "lines+columns").

BTW I tried to minimize the impact to the code. My initial idea was to refactor the type columns into points (using emacs-speak) but it would have a much bigger diffstat. For this reason I opted with this solution, which doesn't change the logic for handling of the output from the previous version of the compiler. Not having color highlight in emacs is a big annoyance.

Thanks!

pmetzger:
Did you close this because you gave up?

chripell:
Never give up hope! ;-)

I had to do some cleaning of my repositories because of corporate policies about which email address I need to associate to my PR/patches. I closed stale PR in the process.

I realize that you folks are busy so I don't want to punt anything at you. I will be maintaining a fork at: https://github.com/chripell/utop (it is not that hard considering the low number of upstream changes) because I think utop+emacs rocks (thanks again for it!). Right now I use ocaml 4.11.1 because Arch Linux is, surprisingly, stuck at it. I haven't testes 4.12 yet. If, some day, you have time to review I am happy to reopen a PR, just ping me.

Thanks!

pmetzger:
If you can give a more thorough explanation of the patches, and resubmit, I'll get someone to review and include the code.

chripell:
Thanks, I will do ASAP. I need also to test on 4.12.

pmetzger:
BTW, do make sure that the description in the merge request explains what you've fixed well enough that a random unfamiliar person will understand what you've done.

pmetzger:
@chripell Just a reminder?

src/lib/uTop.ml
```diff
@@ -126,9 +126,21 @@ let get_ocaml_error_message exn =
...
-      (fun start stop msg -> ((start, stop), msg))
+      (fun start stop msg -> ((start, stop), msg, None))
   with _ ->
-    ((0, 0), str)
+  try
+    Scanf.sscanf
+      str
+      "Line %d, characters %d-%d:\n%[\000-\255]"
+      (fun line start stop msg -> ((start, stop), msg, Some(line, line)))
+  with _ ->
```
rgrinberg, line 136: Can you catch the specific error here instead?

src/lib/uTop.ml
```diff
@@ -126,9 +126,21 @@ let get_ocaml_error_message exn =
...
+  try
+    Scanf.sscanf
+      str
+      "Lines %d-%d, characters %d-%d:\n%[\000-\255]"
+      (fun start_line stop_line start stop msg -> ((start, stop), msg, Some(start_line,stop_line)))
+  with _ ->
```
rgrinberg, line 142: Same here.

src/lib/uTop.ml
```diff
@@ -196,6 +208,7 @@ let discard_formatters pps f =
    +-----------------------------------------------------------------+ *)
 
 type location = int * int
+type lines = int * int
```
rgrinberg, line 211: Can this be a record instead? This will simplify reading the code.

src/lib/uTop_main.ml
```diff
@@ -79,6 +79,39 @@ let index_of_offset src ofs =
...
 let convert_locs str locs = List.map (fun (a, b) -> (index_of_offset str a, index_of_offset str b)) locs
 
+let convert_loc str (a, b) = (index_of_offset str a, index_of_offset str b)
+
+let get_line src line =
+  let rec aux line' ofs skipped =
+    if ofs >= String.length src then
+      ("", 0)
+    else if line' = line then
+      (String.sub src ofs (String.length src - ofs), skipped)
```
rgrinberg, line 89: How come you aren't using `Zed_utf8.sub` and `Zed_utf8.length` here?

src/lib/uTop_main.ml
```diff
@@ -79,6 +79,39 @@ let index_of_offset src ofs =
 
 let convert_locs str locs = List.map (fun (a, b) -> (index_of_offset str a, index_of_offset str b)) locs
 
+let convert_loc str (a, b) = (index_of_offset str a, index_of_offset str b)
```
rgrinberg, line 82: If this function is introduced then `convert_locs` should use it.

pmetzger:
@rgrinberg He closed this, so if you want to resurrect it, you will probably have to do so yourself. 😢

chripell:
Sorry for closing, as I mentioned I needed to change the email address used in my PR because of policies. I was thinking it was OK after 14 months of inactivity to close it.

I was waiting for Arch Linux switching to OCaml 4.12 to do some more day to day testing but it looks it is not going to happen soon. Let me use an opam switch instead and test the patch there.

Thanks for all the comments, I will prepare a *new* PR today addressing them and I will ask for review. Sorry again for the confusion.

src/lib/uTop.ml
chripell: Ack, changing also the untouched for consistency. I used the generic expression because it was used beforehand.

src/lib/uTop.ml
chripell: Ack

src/lib/uTop.ml
chripell: Ack, I used the tuple for consistency with location. 100% agreed record is the right way to go!

src/lib/uTop_main.ml
chripell: It is not clear to me how that can be useful. This function finds the when a new line starts and keeps track of how many characters are before it. I could not pass around skipped and use Zed_utf8.length instead to calculate it. However, that would mean redo a traversal of everything because UTF8.

I agree this code is a bit tricky. IIRC my first implementation was to split the str in lines so you can address easily by line. It was much more memory inefficient but, after all, utop works on small chunks of code. I will submit this piece of code in the new PR and we can move the discussion there. I think the best way to make this more readable is to use the trivial approach to split into lines and have them directly addressable.

src/lib/uTop_main.ml
chripell: Ack

chripell:
I created Issue #349 and PR #350, let's discuss there :-)

pmetzger:
Thank you so much!

chripell:
No problem at all, I will look around the open issues if there are other issues I might help to close (to amortize the learning of utop/utop.el internals ;-) )

Issue #301, opened 2019-10-03
#typeof misparses values containing dots


Octachron:
Trying to get the type of `+.` currently fails with

```OCaml
#typeof "+.";;
Unknown type
```
This is due to the use of `Longident.parse` which only works on a subset of `Longident`s. In the case of `+.`, it misparses the value as `Ldot(Lident "+", "")` and then fails to find such value.

pmetzger:
@Octachron do you have a workaround? (Feel free to patch this if you have a clean one...)

Octachron:
My current plan is to expose a better parser function in compilerlibs, and then use it in utop (so in 4.10 or 4.11). A handwritten parser sounds like too much maintenance work.

pmetzger:
That seems reasonable. In which case, we can leave this open until 4.10 makes it available. Is that your preferred outcome for the moment as well?

Octachron:
Indeed. I essentially opened this issue because I expected the right fix to take some time to propagate.

kandu:
Parse.longident recognizes `(+)` and `(+.)`, but it complains "syntax error" if the brackets are stripped.
Am I doing something wrong?

kandu:
``` ocaml
utop # Parse.longident (Lexing.from_string "+.");;
Exception:
Syntaxerr.Error
 (Syntaxerr.Other
   {Location.loc_start =
     {Lexing.pos_fname = ""; pos_lnum = 1; pos_bol = 0; pos_cnum = 0};
    loc_end = {Lexing.pos_fname = ""; pos_lnum = 1; pos_bol = 0; pos_cnum = 2};
    loc_ghost = false}).
Raised at Parse.wrap in file "parsing/parse.ml", line 65, characters 6-49
Called from Toploop.load_lambda in file "toplevel/toploop.ml", line 212, characters 17-27
```

kandu:
``` ocaml
utop # Parse.longident (Lexing.from_string "+");;
Exception:
Syntaxerr.Error
 (Syntaxerr.Other
   {Location.loc_start =
     {Lexing.pos_fname = ""; pos_lnum = 1; pos_bol = 0; pos_cnum = 0};
    loc_end = {Lexing.pos_fname = ""; pos_lnum = 1; pos_bol = 0; pos_cnum = 1};
    loc_ghost = false}).
Raised at Parse.wrap in file "parsing/parse.ml", line 65, characters 6-49
[...]

PR #320, opened 2020-05-08
compatible with ocaml 4.11

kandu:
* use the exposed parser to parse longident

see also #301

### DIFF
--- a/src/lib/uTop_complete.ml
+++ b/src/lib/uTop_complete.ml
@@ -892,6 +892,14 @@ and find_context_in_quotation = function
    | Completion                                                      |
    +-----------------------------------------------------------------+ *)
 
+#if OCAML_VERSION < (4, 11, 0)
+let longident_parse= Longident.parse
+#else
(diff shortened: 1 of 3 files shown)


src/lib/uTop_complete.ml
```diff
@@ -892,6 +892,14 @@ and find_context_in_quotation = function
    | Completion                                                      |
    +-----------------------------------------------------------------+ *)
 
+#if OCAML_VERSION < (4, 11, 0)
+let longident_parse= Longident.parse
+#else
+let longident_parse str=
+  let lexbuf= Lexing.from_string str in
+  Parse.longident lexbuf
```
Octachron, line 900: I would propose to catch the syntax error here and return an option.

kandu: Below the `complete` function, there is another `complete` function exactly share the same name that wraps and hides the former function with `try .. with`.

So we leave it unchanged?

PR #322, opened 2020-05-08
OCaml 4.11 support

kit-ty-kate:
Fixes https://github.com/ocaml-community/utop/issues/321. cc @kandu

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,8 @@ env:
   - OCAML_VERSION=4.07
   - OCAML_VERSION=4.08
   - OCAML_VERSION=4.09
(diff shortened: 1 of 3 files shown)


kandu:
Thanks! @kit-ty-kate .
As for the warnings, we're going to use the newly exposed parser instead, see #320 
So you may rewrite the history to strip commit `Fix warnings with OCaml 4.11`

kandu:
After this PR is merged, I'll rebase #320 PR to it and merge, so building failure(warnings won't cause failure IIRC) won't last too long.

kandu:
Ah, sorry, the warning fix commit is good. I'll just replace the [%warning] statement with the new parser.

kandu:
Thank you, @kit-ty-kate ^_^

Issue #333, opened 2020-07-19
utop fails to read ~/.config/utop/init.ml


crisperdue:
Documentation states (README.md) that utop will read customizations from ~/.config/utop/init.ml, but
this does not occur, as mentioned also in issue #325. I am running Mac OS 10.15.4 and most recent utop
with ocaml 4.10.0.

Reading customizations as documented would be helpful in setting up consistent behavior of utop
across projects.  My current workaround is to detect the basename of argv.(0) in .ocamlinit and do
customizations for utop based on the result, but this still requires
me to use the API for directives rather than just inserting `#utop_prompt_dummy;;` or other
utop-specific directive, such as `#utop_prompt_simple;;` into a file.

Issue #325, opened 2020-05-13
utop reads .ocamlinit in current directory in preference to .config/utop/init.ml


nilsbecker:
this is the behavior of the bare ocaml toplevel. is this intended also for utop?

i noticed this when starting utop from my home directory, which contains a .ocamlinit. starting utop from any other directory reads the file i expected: .config/utop/init.ml. 
it is possible to move the ocaml toplevel init file to .config/ocaml/init.ml to work around this unexpected behavior of utop.

kandu:
I think @copy is expert in this topic :)

bentxt:
Thanks for that report. I finally understand why my init.ml file is ignore and I cannot the vi input mode (#edit_mode_vi) get to work

PR #338, opened 2020-09-21
Make sure we're not at $HOME before sourcing local .ocamlinit

hyphenrf:
fixes: #333, #325

this `if...else` clause can be bypassed entirely if we don't automatically source `.ocamlinit` in the current project directory and instead pass it as `-init` explicitly.

Not sure what's the default expected ocaml toplevel behaviour though.

### DIFF
--- a/src/lib/uTop_main.ml
+++ b/src/lib/uTop_main.ml
@@ -1479,7 +1479,7 @@ let common_init ~initial_env =
            None
          )
      | None ->
-         if Sys.file_exists ".ocamlinit" && Sys.getcwd () <> Sys.getenv "HOME" then
+         if Sys.file_exists ".ocamlinit" && Sys.getcwd () <> LTerm_resources.home then
            Some ".ocamlinit"
          else
            let xdg_fn = LTerm_resources.xdgbd_file ~loc:LTerm_resources.Config "utop/init.ml" in
--- a/src/lib/uTop_main.ml
+++ b/src/lib/uTop_main.ml
@@ -1479,7 +1479,7 @@ let common_init ~initial_env =
            None
          )
      | None ->
-         if Sys.file_exists ".ocamlinit" then
+         if Sys.file_exists ".ocamlinit" && Sys.getcwd () <> Sys.getenv "HOME" then
            Some ".ocamlinit"
          else
            let xdg_fn = LTerm_resources.xdgbd_file ~loc:LTerm_resources.Config "utop/init.ml" in


src/lib/uTop_main.ml
```diff
@@ -1479,7 +1479,7 @@ let common_init ~initial_env =
            None
          )
      | None ->
-         if Sys.file_exists ".ocamlinit" then
+         if Sys.file_exists ".ocamlinit" && Sys.getcwd () <> Sys.getenv "HOME" then
```
copy, line 1482:
```suggestion
         if Sys.file_exists ".ocamlinit" && Sys.getcwd () <> LTerm_resources.home then
```

copy:
Minor suggestion to stay compatible with Windows and other weird environments, otherwise this looks good.

copy:
This looks good to me now, @kandu could you have a look and merge?

rgrinberg:
Tbh, special casing the home directory like this seems weird to me. The current search strategy makes sense:

- Looks inside cwd
- Look inside XDG
- Look inside $HOME

Special casing $HOME seems strange.

copy:
@rgrinberg I think this change makes sense. If `.ocamlinit` exists in `$HOME`, it's used as a global config file (likely for ocaml, not utop) not a per-project config file, so it should be considered after XDG.

hyphenrf:
> Looks inside cwd
> Look inside XDG
> Look inside $HOME

Intuitively this should be the line of thinking a user expects; as they get more specific (from global config to workspace config), so should utop. running utop from $HOME, that is, when cwd and $HOME are the same thing (e.g. by opening a new shell in a separate window), would violate this principle. It makes sense then to put a special case for when cwd and $HOME are the same thing.. otherwise it expects $HOME to be a workspace config and falls back to the cwd case when even $HOME is meant as a global global config (thanks to legacy homedir assumptions in the ocaml toplevel).

That behaviour _is_ surprising to users as seen in the aforementioned issues.

hyphenrf:
AFAIK ocaml moving to XDG has been a recent change (4.10?) and older compilers don't even recognize XDG's configuration directory

hyphenrf:
maybe an alternative would be to use a different name like .utop for cwd configuration, or to expose an option that doesn't source cwd .ocamlinit unless the user is explicit about it with a flag like --local or something similar..

But I'm definitely not the one with enough confidence to make such a potentially backwards-compat-breaking pr 😅

rgrinberg:
It seems like you are strongly in favour while I'm more ambivalent and think this is meh. The patch in question is tiny, so I don't see why not to merge.

hyphenrf:
I'm willing to put more work in this if you have a better idea in mind ^^

PR #340, opened 2020-11-08
utop.el: Placate byte-compiler

leungbk:


avsm:
Could you describe what problem this PR is solving @leungbk?

pmetzger:
@leungbk Indeed, could you explain the issue this fixes?

pmetzger:
Okay, when I byte compile the file, I do indeed get

```
In utop-mode:
utop.el:1171:19: Warning: reference to free variable ‘company-backends’
    Disable showing Disable logging
utop.el:1171:19: Warning: assignment to free variable ‘company-backends’
    Disable showing Disable logging

In end of data:
utop.el:1238:1: Warning: the following functions are not known to be defined:
    company-begin-backend, company-grab-symbol-cons Disable showing Disable logging
```

But I'm not sure if this patch is the right thing, as we're sort of lying to the byte compiler about what is available. Perhaps the right thing is in fact to conditionally require `company`. In spite of being an old Emacs hand I'm not actually sure.

leungbk:
> But I'm not sure if this patch is the right thing, as we're sort of lying to the byte compiler about what is available. Perhaps the right thing is in fact to conditionally require `company`. In spite of being an old Emacs hand I'm not actually sure.

The `declare-function` invocation does not state whether or not the function is available at byte-compilation time, only where it will be defined at runtime: https://git.savannah.gnu.org/cgit/emacs.git/tree/lisp/emacs-lisp/bytecomp.el?h=a2ae2d28606cef40aa7d076febb0295fcc44be67#n3204

As long as the functions really are defined at runtime, everything works. And the only way that `utop-company-backend` and the stuff buried inside of `(with-eval-after-load 'company...)` can invoke those functions is if we have already loaded `company` at runtime.

rgrinberg:
Probably the right way to fix is to introduce `utop-company` and us `(require 'company)` there. The workaround here is fine as well.

src/top/utop.el
```diff
@@ -1132,6 +1132,8 @@ See https://github.com/diml/utop for configuration information."))
   ;; Load local file variables
   (add-hook 'hack-local-variables-hook 'utop-hack-local-variables))
 
+(defvar company-backends)
```
rgrinberg, line 1135: What happens when the `(defvar company-backends)` in `company.el` is loaded?

leungbk: That will set the value of `company-backends` to the appropriate thing, assuming the user hasn't set it themselves in their config.

The Help page for `defvar` says

> If INITVALUE is missing, SYMBOL's value is not set.

So this `defvar` isn't harmful.

pmetzger:
What should we be doing to get this merged?

rgrinberg:
LGTM. As stated earlier, it's better to introduce `utop-company.el`, but the contributor might have better things to do than work on this.

pmetzger:
Thank you!

PR #347, opened 2021-03-18
Utop correctly restarts when (not utop-edit-command)

dansanduleac:
## Problem

If `utop-edit-command` is nil, then `utop-query-arguments` returns nil.
This breaks `utop-restart` which relies upon it returning the right arguments.

## Solution

`utop-query-arguments` always returns `(utop-arguments)` whether it sets the utop-command or not.


rgrinberg:
Looks good.

PR #350, opened 2021-04-05
Fix color highlight for errors.

### LINKED ISSUE #349: utop doesn't correctly colorize errors in emacs.
At some point the output from ocaml compiler changed and it adds also line information. utop doesn't correctly parse the line part. This is a simple session which shows the problem (I use it also to test the error spanning multiple lines):

```
utop[0]> let f x = x + 1;;
val f : int -> int = <fun>
utop[1]> f 1;;
- : int = 2
utop[2]> f "a";;
Line 1, characters 2-5:
Error: This expression has type string but an expression was expected of type
         int
utop[3]> f "a
b";;
Lines 1-2, characters 2-2:
Error: This expression has type string but an expression was expected of type
         int
```

I expect to have the string colored in red, instead I just get the position of the error. This makes really difficult to understand where the error is for long expressions.

I will propose a PR shortly.

Thanks!

chripell:
The output from the compiler changed and now mentions the line number
(non only the charters, which actually becomes columns). This patch
introduces the idea of a range of lines spanned by an error and
correctly marks the faulty region.

I think a better solution would be to introduce a "point" record of
the form {line:int;column:int}. However, such a change would
reverberate through all the source code and so it is much more
invasive. I can prepare such a patch if desired, after this one, which
fixes the regression.

This was tested with OCaml 4.11.2 and 4.12.

Fixes #349

### DIFF
--- a/src/lib/uTop.ml
+++ b/src/lib/uTop.ml
@@ -104,6 +104,15 @@ let keywords = ref (String_set.of_list default_keywords)
 let add_keyword kwd = keywords := String_set.add kwd !keywords
 
 (* +-----------------------------------------------------------------+
+   | Span of Lines                                                   |
+   +-----------------------------------------------------------------+ *)
+
+type lines = {
+  start: int;
+  stop:  int;
+}
+
+(* +-----------------------------------------------------------------+
    | Error reporting                                                 |
    +-----------------------------------------------------------------+ *)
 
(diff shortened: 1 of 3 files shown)


pmetzger:
@rgrinberg Care to review and/or merge?

chripell:
I've checked the CI errors. The Mac version seems broken for unrelated reasons (I suspect the timeout is a result of something happening beforehand, homebrew update itself is taking forever; however, it is years that I don't touch a Mac and even more since I used homebrew):
```
make world.opt
The job exceeded the maximum time limit for jobs, and has been terminated.
```

The one Linux failing refers to a non-existent switch:
```
+opam switch ocaml-variants.4.11.0+trunk
[ERROR] No switch ocaml-variants.4.11.0+trunk is currently installed. Did you
        mean 'opam switch create ocaml-variants.4.11.0+trunk'?
        Installed switches are:
+opam switch create --repo=default,beta ocaml-variants.4.11.0+trunk --packages=ocaml-variants.4.11.0+trunk
[ERROR] ocaml-variants = 4.11.0+trunk no matching version
The command "bash -ex .travis-opam.sh" exited with 5.
```
I would suggest that this should be moved to 4.12, I can send a patch for https://github.com/ocaml-community/utop/blob/master/.travis.yml if you want.

All other CIs are green, so the "all checks have failed" is kind of a misleading message.

pmetzger:
We really should be moving from Travis to Github Actions of course for our CI.

rgrinberg:
LGTM

Issue #343, opened 2021-01-24
utop-company-backend trying to complete numbers


rmunderwood:
Whenever I type any number into utop with emacs, for some reason the utop-company-backend tries to complete it and lists every library/completion, it is a bit of an annoyance. Is there any solution/workaround to this or have I just set up my environment incorrectly? Thanks
![Screenshot at 2021-01-24 15-20-16](https://user-images.githubusercontent.com/76738501/105634858-fcb7c600-5e57-11eb-98cd-2403aa1ff5a2.png)

chripell:
It looks like it is the normal behavior for utop. You can verify it by running the CLI version of `utop`, you have *all* completion by default. It can be worked out by doing the filtering in the elisp code, here a proof of concept patch (I haven't tested it a lot!):

```
--- utop-oldy.el	2021-04-05 08:51:50.437976506 +0100
+++ utop.el	2021-04-05 14:52:50.240568868 +0100
@@ -175,6 +175,9 @@
 (defvar utop-completion nil
   "Current completion.")
 
+(defvar utop-completion-prefixes nil
+  "Prefixes for current completion.")
+
 (defvar utop-inhibit-check nil
   "When set to a non-nil value, always insert text, even if it is
 before the end of prompt.")
@@ -645,7 +648,10 @@
        (setq utop-completion nil))
       ;; A new possible completion
       ("completion"
-       (push argument utop-completion))
+       (catch 'done
+	 (dolist (prefix utop-completion-prefixes)
+	   (if (string-prefix-p prefix argument)
+	       (progn (push argument utop-completion) (throw 'done t)) nil))))
       ;; End of completion
       ("completion-stop"
        (utop-set-state 'edit)
@@ -757,6 +763,9 @@
      (if (utop--supports-company)
          "complete-company:\n"
        "complete:\n"))
+    ;; Keep track of the prefixes, so we can avoid returning
+    ;; completion which don't have a match.
+    (setq utop-completion-prefixes lines)
     (dolist (line lines)
       ;; Send the line
       (utop-send-string (concat "data:" line "\n")))
```

I think that filtering is the right behavior, so I will prepare a PR to check if the utop authors agree.

rgrinberg:
@chripell that looks ok to me.

pmetzger:
@chripell Would you like to do a pull request?

chripell:
Sent PR, sorry for the delay because of a busy week.

PR #352, opened 2021-05-22
Fix emacs completion returning bogus candidates.

chripell:
In standalone mode, utop returns all the possible completions when no
one matches what the user has typed. This is surprising in emacs
because of the different UI/UX. For this reason, we keep a list of
prefixes for which completion has started and return only the
candidates which match the prefixes.

Fixes: #343

### DIFF
--- a/src/top/utop.el
+++ b/src/top/utop.el
@@ -175,6 +175,9 @@ This hook is only run if exiting actually kills the buffer."
 (defvar utop-completion nil
   "Current completion.")
 
+(defvar utop-completion-prefixes nil
+  "Prefixes for current completion.")
+
 (defvar utop-inhibit-check nil
   "When set to a non-nil value, always insert text, even if it is
 before the end of prompt.")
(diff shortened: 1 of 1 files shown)


pmetzger:
@rgrinberg I presume we should merge this?

src/top/utop.el
```diff
@@ -645,7 +648,10 @@ it is started."
        (setq utop-completion nil))
       ;; A new possible completion
       ("completion"
-       (push argument utop-completion))
+       (catch 'done
+	 (dolist (prefix utop-completion-prefixes)
+	   (if (string-prefix-p prefix argument)
+	       (progn (push argument utop-completion) (throw 'done t)) nil))))
```
rgrinberg, line 654: Why do you need the `nil` at the end?

src/top/utop.el
```diff
@@ -645,7 +648,10 @@ it is started."
        (setq utop-completion nil))
       ;; A new possible completion
       ("completion"
-       (push argument utop-completion))
+       (catch 'done
+	 (dolist (prefix utop-completion-prefixes)
```
rgrinberg, line 652: The indentation seems off. `(dolist` should be indented because it's inside `catch`

src/top/utop.el
chripell: Sorry, leftover from a previous version which had also an else branche. Fixed.

src/top/utop.el
chripell: Fixed.

chripell:
PTAL

pmetzger:
@rgrinberg Let me know if this is now mergeable.

pmetzger:
@rgrinberg ?

rgrinberg:
Merged with a minor modifications. Thanks.

chripell:
Thanks!

PR #353, opened 2021-05-25
Add support for OCaml 4.13

kit-ty-kate:
Changes came from https://github.com/ocaml/ocaml/pull/10124 as well as https://github.com/ocaml/ocaml/pull/10207

Only 300dd7e is strictly necessary to build utop on OCaml 4.13.


rgrinberg:
Thanks Kate.

PR #362, opened 2021-11-24
Move histfile to the system's "state" folder

exorcist365:


src/lib/uTop.ml
```diff
@@ -15,6 +15,8 @@ open LTerm_text
 open LTerm_geom
 open LTerm_style
 
+module M = Directories.Base_dirs ()
```
zapashcanon, line 18: I don't think that's the best way to do it, it would probably be better like this:

```ml
module M = Directories.Project_dirs (struct
    let qualifier = "org"
    let organization = "ocaml-community"
    let application = "utop"
end)
```
And you'll have to create the directory yourself if it doesn't exist.

src/lib/uTop.ml
```diff
@@ -33,7 +35,10 @@ let version = "%%VERSION%%"
    +-----------------------------------------------------------------+ *)
 
 let history = LTerm_history.create []
-let history_file_name = ref (Some (Filename.concat LTerm_resources.home ".utop-history"))
+let unopt = function
+        | None -> "None"
+        | Some(v) -> v
+let history_file_name = ref(Some(Filename.concat (unopt(M.cache_dir)) ".utop_history"))
```
zapashcanon, line 41: Using `"None"` as a directory name is probably not what should be done. You should either fail if `M.cache_dire` is `None` (this happens only if there's no way to guess the user's home and directories try quite hard to guess it, so that almost never happen in practice, I've never seen it actually) or just use the current directory and print a warning or whatever.

src/lib/uTop.ml
exorcist365: Alright, I will add that, I was wondering if this was how I had to do it but I didn't know what to put there.

src/lib/uTop.ml
```diff
@@ -35,8 +40,8 @@ let version = "%%VERSION%%"
    +-----------------------------------------------------------------+ *)
 
 let history = LTerm_history.create []
-let unopt = function
-        | None -> "None"
+let rec unopt = function
+        | None -> unopt(H.home_dir)
```
zapashcanon, line 44: if `H.home_dir` is `None`, this will loop indefinitely...

exorcist365: It's never going to happen that H.home_dir is none.

exorcist365: But I'll change it, sure.

exorcist365:
I'm not exactly sure how to create the directory if it doesn't exist, I'm still new to OCaml and not yet familiar with most libraries.

src/lib/uTop.ml
```diff
@@ -33,7 +39,12 @@ let version = "%%VERSION%%"
    +-----------------------------------------------------------------+ *)
 
 let history = LTerm_history.create []
-let history_file_name = ref (Some (Filename.concat LTerm_resources.home ".utop-history"))
+
+let unopt = function
+        | Some(v) -> v
+        | None -> "~/"
```
zapashcanon, line 45: I still believe this is not a good idea.

Here's a solution :

```ml
let history_file_name = ref (Option.map (fun cache_dir -> Filename.concat cache_dir "history") M.cache_dir)
```

(and remove your `unopt` function)

exorcist365: But then again, what if M.cache_dir is None?

zapashcanon: It's [already handled](https://github.com/ocaml-community/utop/blob/560617c64f544dba7fe5b6f1a2ad1a963b3d36e4/src/lib/uTop_main.ml#L47-L49) and it won't happen in practice.

exorcist365: Alright then.

src/lib/uTop.ml
```diff
@@ -41,10 +41,13 @@ let version = "%%VERSION%%"
...
-        | Some(v) -> v
-        | None -> "~/"
+    | None -> ""
+    | Some(v) -> v
 
-let history_file_name = ref(Some(Filename.concat (unopt M.cache_dir) ".utop_history"))
+let () =
+    try ignore(Sys.is_directory (unopt M.cache_dir)) with Sys_error(_) -> Sys.mkdir (unopt M.cache_dir) 755
+
+let history_file_name = ref (Option.map (fun cache_dir -> Filename.concat cache_dir "history") M.cache_dir)
```
zapashcanon, line 50:
```ml
let () = Option.iter (fun cache_file ->
    let dir = Filename.dirname cache_file in
    Sys.mkdir dir 0o600; (* This isn't equivalent to `mkdir -p` which is what should be implemented *)
  ) !history_file_name
```

[Here](https://github.com/dbuenzli/bos/blob/master/src/bos_os_dir.ml#L17) you have an equivalent of `mkdir -p` in OCaml.

zapashcanon:
I'm not sure that adding `Bos` as a dependency is something we want, but I'm not the one who can make the decision here.

If it's OK to have it, then, you should probably match on the result and print a warning/fail if the dir hasn't been created.

If it's not, you could just reimplement the part of `Bos.OS.Dir.create` that is relevent here.

pmetzger:
Adding more dependencies for something very small is probably not optimal, no.

lindig:
The code below can be used to create a directory, including parent directories. If this is what you need.

```
(** create a directory but doesn't raise an exception if the directory already exist *)
let mkdir_safe dir perm =
  try Unix.mkdir dir perm with Unix.Unix_error (Unix.EEXIST, _, _) -> ()

(** create a directory, and create parent if doesn't exist *)
let mkdir_rec dir perm =
  let rec p_mkdir dir =
    let p_name = Filename.dirname dir in
    if p_name <> "/" && p_name <> "."
    then p_mkdir p_name;
    mkdir_safe dir perm in
  p_mkdir dir
```

src/lib/uTop.ml
```diff
@@ -32,8 +38,29 @@ let version = "%%VERSION%%"
...
+(** create a directory but doesn't raise an exception if the directory already exist *)
+let mkdir_safe dir perm =
+    try Unix.mkdir dir perm with Unix.Unix_error (Unix.EEXIST, _, _) -> ()
+
+(** create a directory, and create parent if doesn't exist *)
+let mkdir dir perm =
+    let rec p_mkdir dir =
+        let p_name = Filename.dirname dir in
+        if p_name <> "/" && p_name <> "."
+    then p_mkdir p_name;
```
lindig, line 50: Indentation is misleading here. In general, I would suggest to move to OCamlformat to avoid discussions around indentation.

src/lib/uTop.ml
```diff
@@ -32,8 +38,29 @@ let version = "%%VERSION%%"
...
+    p_mkdir dir
+
 let history = LTerm_history.create []
-let history_file_name = ref (Some (Filename.concat LTerm_resources.home ".utop-history"))
+
+let history_file_name = ref (Option.map (fun cache_dir -> Filename.concat cache_dir "history") M.cache_dir)
+
+let () = Option.iter (fun cache_file ->
+    let dir = Filename.dirname cache_file in
+    if not (Sys.file_exists dir) then begin
```
lindig, line 60: The `Sys.file_exists` is not required because `mkdir` does not fail if the directory exists.

src/lib/uTop.ml
```diff
@@ -15,6 +15,12 @@ open LTerm_text
 open LTerm_geom
 open LTerm_style
 
+module M = Directories.Project_dirs (struct
```
lindig, line 18: `M` is not a good name - because what would it stand for? I assume this module provides `M.cache_dir`. A better name would be `D`, `Dirs`, or `Project`.

zapashcanon: What about `Project_dirs` ?

lindig: That would work as well.

src/lib/uTop.ml
```diff
@@ -32,8 +38,29 @@ let version = "%%VERSION%%"
...
+    p_mkdir dir
+
 let history = LTerm_history.create []
-let history_file_name = ref (Some (Filename.concat LTerm_resources.home ".utop-history"))
+
+let history_file_name = ref (Option.map (fun cache_dir -> Filename.concat cache_dir "history") Project_dirs.cache_dir)
+
+let () = Option.iter (fun cache_file ->
+    let dir = Filename.dirname cache_file in
+    begin
```
zapashcanon, line 60: This `begin` and the `end` are not necessary. It's the indentation in `mkdir` that is confusing.

exorcist365: Yeah, i fixed it like 2 minutes later.

src/lib/uTop.ml
```diff
@@ -32,8 +38,26 @@ let version = "%%VERSION%%"
    | History                                                         |
    +-----------------------------------------------------------------+ *)
 
+(** create a directory but doesn't raise an exception if the directory already exist *)
+let mkdir_safe dir perm =
+    try Unix.mkdir dir perm with Unix.Unix_error (Unix.EEXIST, _, _) -> ()
+
+(** create a directory, and create parent if doesn't exist *)
+let mkdir dir perm =
```
lindig, line 46: This has still confusing and/or unconventional indentation. It should be more like the code below. In particular, you want to avoid to increase indentation after `let .. in` because that would force your code quickly to the right. There also appears to be a mix of how much indentation you use for each level.

```
let mkdir dir perm =
  let rec p_mkdir dir =
    let p_name = Filename.dirname dir in
    if p_name <> "/" && p_name <> "." then p_mkdir p_name;
    mkdir_safe dir perm
  in
  p_mkdir dir

let history = LTerm_history.create []

let history_file_name =
  ref
    (Option.map
       (fun cache_dir -> Filename.concat cache_dir "history")
       Project_dirs.cache_dir)

let () =
  Option.iter
    (fun cache_file ->
      let dir = Filename.dirname cache_file in
      mkdir dir 0o755)
    !history_file_name

```

Firobe:
Thanks for making this PR ! This is a detail but shouldn't the file be in XDG_STATE_HOME instead of XDG_CACHE_HOME ? At least that's what the specification (https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) mentions.

zapashcanon:
Oh, I didn't know about it thanks ! It's been introduced in the latest basedir spec. I'll add it to directories. :)

zapashcanon:
Done, should be available soon, see ocaml/opam-repository#20409.

src/lib/uTop.ml
```diff
@@ -32,8 +38,34 @@ let version = "%%VERSION%%"
...
+  p_mkdir dir
+
 let history = LTerm_history.create []
-let history_file_name = ref (Some (Filename.concat LTerm_resources.home ".utop-history"))
+
+let history_file_name =
+  ref
+    (Option.map
+       (fun cache_dir -> Filename.concat cache_dir "history")
+       Project_dirs.cache_dir)
```
zapashcanon, line 60:
```ml
let history_file_name =
  ref
    (Option.map
       (fun state_dir -> Filename.concat state_dir "history")
       Project_dirs.state_dir)
```

utop.opam
```diff
@@ -12,6 +12,7 @@ depends: [
   "ocamlfind" {>= "1.7.2"}
   "lambda-term" {>= "3.1.0" & < "4.0"}
   "lwt"
+  "directories"
```
zapashcanon, line 15:
```yaml
  "directories" {>= "0.5"}
```

zapashcanon:
Thanks @exorcist365. I believe the PR can be merged but I don't have the rights to do it...

emillon:
Hi! I feel like adding a transitive dependency to `ctypes` on windows (through `directories`) is a bit heavy. Is there a way to have native bindings to the corresponding APIs?

zapashcanon:
Note that some effort has been made not to depend on `ctypes.foreign` and `libffi` but only `ctypes.stubs` and `ctypes`. Regarding native bindings, I personally won't go into the trouble of it.

emillon:
Sure, I understand. To give you some context, we're making sure that a selection of packages (including utop) are available as soon as new versions of the ocaml compiler are released. Adding ctypes on windows to the mix seems difficult but I'll check with the rest of the team.

emillon:
It looks like `xdg` is going to provide this kind of support on windows without the ctypes dependency: https://github.com/ocaml/dune/pull/5943
If this works in the context of this PR, I think that this would be preferable in terms of dependencies.

zapashcanon:
Well, in the current state the PR doesn't look correct to me (I commented there), but if it ends up providing a proper way to get the cache directory without ctypes it's great.

emillon:
I'm not willing to take a dependency on ctypes at the moment, but please reopen if you're willing to implement this on top of `xdg` or a similar lighter alternative. thanks!

PR #371, opened 2022-03-28
Remove -safe-string for OCaml 5. See ocaml/ocaml#10893

favonia:
ocaml/ocaml#10893 removed the flag `unsafe_string` in the upcoming OCaml 5.


kit-ty-kate (approved):

kit-ty-kate:
Thanks a lot!

PR #373, opened 2022-04-16
Adapt to new mechanism for loading topdirs.cmi

dra27:
Provides compatibility with the proposed change in ocaml/ocaml#11199


kit-ty-kate:
I don't think this was meant to be merged now as the function does not exist yet

emillon:
I think we can keep as is and the testing we'll do for 5.0.0 will find if this was the right function :)

dra27:
I just merged the upstream OCaml PR, so the merge here was simply ahead of the curve 🙂

PR #376, opened 2022-05-29
Support OCaml 5, again

favonia:
This PR makes utop compile again with OCaml 5. The main change is to explicitly load packages from OCaml otherlibs, following the spirit of https://github.com/ocaml/ocaml/pull/11198.

**API changes:** I introduced `UTop.otherlibs` and `UTop.load_otherlibs` and an optional argument to `UTop.set_load_path` to control whether that function should load packages from OCaml otherlibs automatically.

**Caveats:** I did not test the new code with older versions of OCaml. It is possible that the explicit list of packages (`UTop.otherlibs`) depends on the version of the OCaml compiler. I don't have the energy to check older versions of OCaml and please feel free to update this PR.

### DIFF
--- a/src/lib/uTop.ml
+++ b/src/lib/uTop.ml
@@ -850,10 +850,23 @@ let () =
   Topdirs.dir_directory (Findlib.package_directory "utop")
 
  (* +-----------------------------------------------------------------+
-   | Compiler-libs re-exports                                        |
-   +-----------------------------------------------------------------+ *)
+    | Compiler-libs re-exports and helpers                            |
+    +-----------------------------------------------------------------+ *)
 
-#if OCAML_VERSION >= (4, 08, 0)
+let otherlibs = ["dynlink"; "runtime_events"; "str"; "threads"; "unix"]
+
+(* Explicitly add otherlibs as many components live in different directories since OCaml 5.0. *)
+let load_otherlibs () =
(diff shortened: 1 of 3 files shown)


pmetzger:
Side note: we really need a 5.0 pre-release CI test.

dra27:
Apologies - during the review of 11198 the new argument ceased to be optional, and I didn’t register that would affect utop.

Out of curiosity, why is utop going to such lengths here? Once ocamlfind is released with the update for the layout (the PR’s already open) `#require "unix";;` will do the right thing.

favonia:
@dra27 The main reason is that I want something that is working for me _now._ However, I could comment on why this might not look terribly wrong: `utop` will automatically load these packages the moment the user types, for example, `Unix.` (even before pressing the enter key) for auto-completion; the user would not have the chance to input `#require` before `utop` automatically loads those packages. Any unexpected alert would mess up the user interface because `utop` assumes it has the entire screen under its control. Either `utop` has to change its expectation or somehow silence the alert. I made a change that I thought is easy, local, and working and, just to clarify, I am not one of the `utop` developers. :sweat_smile:

favonia:
@dra27 I realized that my local ocamlfind was already patched because I have been using the alpha repository. So, no, the UI issue definitely cannot be solved by just patching `ocamlfind`, for the reasons described above. `utop` has to change how it uses compilerlib (for example adopting the hack here), change how it controls the screen, or do something creative.

PR #378, opened 2022-06-06
Add toplevel_printer support for functors

metanivek:
This PR extends #269 by adding support for annotations with functors.

This issue was discovered while attempting to add pretty-printers for Irmin (https://github.com/mirage/irmin/pull/1839). 

I centralized the code for these annotations into a module named `Autoprinter`. It is a combination of code from a proof-of-concept created for this issue (https://github.com/let-def/autoprinter) and the existing code for automatically installing printers.

I have tested this for each conditional compiler check in the code (4.04, 4.06, 4.08, 4.09, 4.10) using both the Irmin PR (when compatible) above and a simple test project (https://github.com/metanivek/pp_utop_testing). Additionally, I noticed that `ocp-indent` didn't format the code well, so I tried to match the existing indentation style.

Feedback and comments welcome!

### DIFF
--- a/src/lib/uTop_main.ml
+++ b/src/lib/uTop_main.ml
@@ -309,6 +309,126 @@ class read_phrase ~term = object(self)
 end
 
 (* +-----------------------------------------------------------------+
+   | Handling of [@@toplevel_printer] attributes                     |
+   +-----------------------------------------------------------------+ *)
+
+module Autoprinter = struct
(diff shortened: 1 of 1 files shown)


metanivek:
Thanks for the review. I've rebased and updated the ifdefs. I'm very happy to have fewer compiler checks in the code. :smile:

PR #382, opened 2022-07-04
Compatibility with zed 3.2.0

emillon:
This updates utop to use the uu packages instead of camomile, following ocaml-community/zed#46 and ocaml-community/lambda-term#109.

### DIFF
--- a/src/lib/dune
+++ b/src/lib/dune
@@ -4,7 +4,7 @@
  (wrapped false)
  (flags :standard -safe-string)
  (modes byte)
- (libraries compiler-libs.toplevel findlib.top lambda-term threads)
+ (libraries compiler-libs.toplevel findlib.top lambda-term threads zed)
  (preprocess
   (action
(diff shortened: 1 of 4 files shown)


emillon:
This requires https://github.com/ocaml-community/lambda-term/pull/109 to be merged first. There's probably a way to be compatible with both camomile and zed but this is fine like this.

emillon:
(the nix builds are happy but they're just running `dune runtest` which always succeed, need to fix that separately)

kit-ty-kate (approved):

kit-ty-kate:
I’ve just enabled ocaml-ci here so we can see the proper result

olafhering:
The `README.md` and `nix` files still refer to camomile.

emillon:
I added #383 to get rid of the duplication in the README. I can't update the nix lockfiles so once ocamlci works well I think I'll remove the corresponding builds.

PR #388, opened 2022-07-16
[Fix #358] Fix code completion for module identifiers in utop.el

### LINKED ISSUE #358: utop in Emacs couldn't auto-complete after dot
Emacs fails to auto-complete inputs such as "List.m TAB".
IMHO, the string-prefix-p test is the reason, because the prefix "List." is not shown in the completion data.
```
$ utop -emacs
protocol-version:1
phrase-terminator:;;
stdout:Welcome to utop version 2.7.0 (using OCaml version 4.12.0)!
stdout:
prompt:
complete-company:
data:Li
end:
completion-start:
completion:List
completion:ListLabels
completion-stop:
complete-company:
data:List.m
end:
completion-start:
completion:map
completion:map2
completion:mapi
completion:mem
completion:mem_assoc
completion:mem_assq
completion:memq
completion:merge
completion-stop:
```
https://github.com/ocaml-community/utop/blob/bac3946079a98df00f31656dc10c6c9f1a8bc422/src/top/utop.el#L650-L655

bbatsov:
Without those changes you won't get any completion for things like
"List.m". I'm not very familiar with the code, but it seems that the
backend returns the completion candidates without the module prefix
which caused the problem.

I've opted to fix this by handling prefixes with a "." in them specially
and this works fine in the limited testing I conducted locally.

### DIFF
--- a/src/top/utop.el
+++ b/src/top/utop.el
@@ -243,7 +243,7 @@ backend")
 
 (defun utop--supports-company ()
   (and
-   ;; version< only works on version numbers 
+   ;; version< only works on version numbers
    (condition-case nil
        (version<= "1" utop-protocol-version)
      (error t))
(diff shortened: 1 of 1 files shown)


idurand:
Hi,
After a few months with doom Emacs which drove me crazy (but gave me auto-completion in utop Emacs),
and after my laptop was stolen I am back with Emacs but still don't find a fix on utop.el so that
completion works inside packages:
for instance
List.<TAB>
Is this possible with just tuareg and utop?
Was this problem ever solved?
Regards

PR #391, opened 2022-07-18
Add commands to switch to and from a utop buffer

bbatsov:
Those are extremely common in Emacs modes and allow people to
quickly jump to a REPL and back to the last source buffer they
were editing, while using the same keybinding.

I've modeled the implementation here after that of inf-clojure (a similar
mode written by me). I've also implemented a mode menu for
utop-minor-mode, so it's easier for the users to discover its
functionality.

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,13 +218,14 @@ You can start utop inside Emacs with: `M-x utop`.
 
 `utop.el` also ships with a minor mode that has the following key-bindings:
 
-| key-binding | function          | Description                  |
-|-------------|-------------------|------------------------------|
-| C-c C-s     | utop              | Start a utop buffer          |
-| C-x C-e     | utop-eval-phrase  | Evaluate the current phrase  |
-| C-x C-r     | utop-eval-region  | Evaluate the selected region |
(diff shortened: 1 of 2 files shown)


rgrinberg:
Is there a way to add tests for this stuff?

I don't think many people know elisp well enough to review your changes for correctness.

bbatsov:
@rgrinberg It's possible, but tests for something like this require a setup that's much more complicated than the two commands that were added (as you need to have several source buffers and you have to switch between them to test the recency, an `utop` buffer and `utop-minor-mode` running in the source buffers). That's why I rarely bother to automate the testing of such interactions and I usually only do some unit tests when the ROI is better.  

> I don't think many people know elisp well enough to review your changes for correctness.

That's a common problem unfortunately. :-)

bbatsov:
(I do test all the changes I make locally before submitting them, though, so I'm reasonably certain that this works; it's also isolated from the rest of the functionality, so it can't really break anything either)

Btw, down the road I can at least add some linting for utop.el to the CI, which would help discover some basic errors.

PR #393, opened 2022-07-19
Bump the compat to 4.08+

emillon:
Utop supports a wide range of OCaml versions, but this is getting more difficult to support. In addition, some of the dependencies we use do not support 4.03-4.07, so utop can not be built on these versions.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,8 @@
+Unreleased
+----------
+
+* Bump the compatibility to 4.08+ (#393 @emillon)
+
 2.10.0 (2022-10-06)
 ------------------
 
--- a/src/lib/uTop.ml
+++ b/src/lib/uTop.ml
@@ -204,10 +204,7 @@ let discard_formatters pps f =
   (* Output functions. *)
   let out_functions = {
     Format.out_string = (fun _ _ _ -> ()); out_flush = ignore;
-    out_newline = ignore; out_spaces = ignore
-#if OCAML_VERSION >= (4, 06, 0)
-      ; out_indent = ignore
-#endif
+    out_newline = ignore; out_spaces = ignore ; out_indent = ignore
   } in
   (* Replace formatter functions. *)
   List.iter (fun pp -> Format.pp_set_formatter_out_functions pp out_functions) pps;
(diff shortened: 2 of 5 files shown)


rgrinberg (approved):

CHANGES.md
```diff
@@ -1,3 +1,8 @@
+Unreleased
+----------
+
+* Bump the compatibility to 4.08+ (#393 @emillon)
```
bbatsov, line 4: Btw, I know you've been doing the changelog like this for a while, but it's a bit better if you include actual links here (especially for the tickets). It's much easier to just clink on a link instead of having to type in the URL to an issue.

bbatsov: To be a bit more specific - I'm suggesting doing something like this instead https://github.com/rubocop/rubocop/blob/master/CHANGELOG.md

emillon: (to clarify my point: in make sense to do that but there's much more to do first in utop)

bbatsov: > I appreciate the suggestion, but this is too much work to do by hand every time (in rubocop, [this is generated](https://github.com/rubocop/rubocop/blob/master/tasks/changelog.rb)). I agree that the end result is nice, but I prefer to keep the bar low so as not to abandon the changelog.

Just to clarify - in RuboCop it's not generated either. Just individual entries that are manually written are put together by a rake task at release time. The purpose of the generation is to avoid conflicts in the changelog, not to make it easier to write changelog entries. 

Anyways, I see your point, but in general there's little difference to copying a link over just writing an issue number. Perhaps I'm just way too used to this to notice at this point. :-)

emillon: The link to the usernames are generated as well. They're not in the source: https://github.com/rubocop/rubocop/blob/2f8d8c7d9fe62b32260959a73a9a231cd05f3958/changelog/new_add_new_style_empty_heredoc_cop.md
But yeah, I thought that the links to issues were generated too, and they're not indeed.

bbatsov: The links to the usernames are generated because people kept forgetting to add them to the bottom of the changelog (as they had to be added just once to be references easier later). 10 years of working on RuboCop made me an expert on every changelog issue that might possibly exist. 😆 I guess that's why I developed some unhealthy obsession with changelogs at the end of the day.

PR #394, opened 2022-07-19
Remove nix build

emillon:
This build does not work, we don't have an easy way to keep the lockfiles updated, and we now have ocaml-ci covering the builds.


README.md
```diff
@@ -1,5 +1,3 @@
-[![Build Status](https://github.com/ocaml-community/utop/workflows/Test%20in%20Nix/badge.svg)](https://github.com/ocaml-community/utop/actions/workflows/test-in-nix.yml)
```
bbatsov, line 1: Why don't you update this badge to point to the ocaml-ci you mentioned?

PR #395, opened 2022-07-19
A bunch of small improvements to utop.el

bbatsov:
This mostly polishes my previous work. I noticed a couple of small oversights on my end and I also made a couple of random improvements. Most notably - I identified a big chunk of dead code (that never worked in the first place) and opted to remove it.


bbatsov:
Probably this is the last change I'll do for a while, as I've been spending more time fixing the Emacs packages for OCaml than writing OCaml. :D But hacking on Emacs packages is so much fun. :-)

bbatsov:
It'd be nice if we also cut a stable release of utop.el afterwards - e.g. 1.12 or 2.0. I'll leave this decision up to you.

src/top/utop.el
```diff
@@ -355,51 +360,6 @@ it is started."
       (utop-send-string (concat "data:" line "\n")))
     (utop-send-string "end:\n")))
 
-(defun utop-last-type ()
```
rgrinberg, line 358: Supposedly, this would b e useful to insert inferred type annotations from the toplevel into the source. However, there's now better ways to do this (merlin, lsp)

bbatsov:
@rgrinberg What would like to see squashed? For me currently all distinct changes in separate commits. Perhaps only the `when-let` commit should go away.

rgrinberg:
All the grammar fixes can be a single commit.

PR #402, opened 2022-10-26
utop.el: always insert phrase terminator at the end-ish of input

leungbk:
When point is at the letter "n" in

utop[1]> let thi[n]g = 42

and the user presses C-j, it is always because they want to terminate
the expression at the end-ish of the input.

The unpatched behavior did not move point, and inserted two semicolons
in the middle of the expression.

### DIFF
--- a/src/top/utop.el
+++ b/src/top/utop.el
@@ -838,7 +838,7 @@ If ADD-TO-HISTORY is t then the input will be added to history."
   "Go to the beginning of line or to the end of the prompt."
   (interactive)
   (with-current-buffer utop-buffer-name
-    (if (= (point-at-bol) utop-prompt-min)
+    (if (= (line-beginning-position) utop-prompt-min)
         (goto-char utop-prompt-max)
       (move-beginning-of-line 1))))
 
--- a/src/top/utop.el
+++ b/src/top/utop.el
@@ -649,8 +649,8 @@ If ALLOW-INCOMPLETE is non-nil and the phrase is not terminated,
 then a newline character will be inserted and edition will
 continue.
 
-If AUTO-END is non-nill then ALLOW-INCOMPLETE is ignored and a
-phrase terminator (;; or ; if using revised syntax) will be
+If AUTO-END is non-nill then ALLOW-INCOMPLETE is ignored and the
+phrase terminator (;;) will be
 automatically inserted by utop.
 
 If ADD-TO-HISTORY is t then the input will be added to history."
--- a/src/top/utop.el
+++ b/src/top/utop.el
@@ -649,7 +649,7 @@ If ALLOW-INCOMPLETE is non-nil and the phrase is not terminated,
 then a newline character will be inserted and edition will
 continue.
 
-If AUTO-END is non-nill then ALLOW-INCOMPLETE is ignored and the
+If AUTO-END is non-nil then ALLOW-INCOMPLETE is ignored and the
 phrase terminator (;;) will be
 automatically inserted by utop.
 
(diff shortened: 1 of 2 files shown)


leungbk:
cc @rgrinberg

rgrinberg:
@bbatsov can you review please?

bbatsov:
Will do!

src/top/utop.el
```diff
@@ -842,7 +842,7 @@ If ADD-TO-HISTORY is t then the input will be added to history."
   "Go to the beginning of line or to the end of the prompt."
   (interactive)
   (with-current-buffer utop-buffer-name
-    (if (= (point-at-bol) utop-prompt-min)
```
bbatsov, line 845: Is this deprecated or what?

src/top/utop.el
```diff
@@ -488,12 +488,12 @@ it is started."
   "Insert the phrase terminator at the end of buffer."
   ;; Search the longest suffix of the input which is a prefix of the
   ;; phrase terminator
-  (let* ((end (point-max))
-         (pos (max utop-prompt-max (- end (length utop-phrase-terminator)))))
-    (while (not (string-prefix-p (buffer-substring-no-properties pos end) utop-phrase-terminator))
-      (setq pos (1+ pos)))
+  (let ((end (point-max)))
+    (goto-char (max utop-prompt-max (- end (length utop-phrase-terminator))))
+    (while (not (string-prefix-p (buffer-substring-no-properties (point) end) utop-phrase-terminator))
```
bbatsov, line 493: Frankly, this feel super wasteful. We're moving char by char to check if the last two characters are `;;`? Seems to me it'd be much easier to check the suffix of the string and insert the `;;` if not present. You can use either `string-suffix-p`, regex directly or `looking-back`. This would simplify this convoluted code a lot IMO.

bbatsov:
Btw, that's how - tuareg does exactly the same thing https://github.com/ocaml/tuareg/blob/master/tuareg.el#L3788

Might not be a bad idea to just do the same, as it's obviously simpler.

src/top/utop.el
leungbk: I agree it's wasteful, but I was trying to preserve the semantics of the old code, which attempted to account for the case where the phrase terminator was something else.

From a quick glance at the OCaml files, the phrase terminator can be a single semicolon (https://github.com/diml/utop/blob/c68acd1ff3c350f33ef8c270fa3de236f832f5e1/src/lib/uTop.mli#L104-L106) when using "revised" syntax. However, grepping for `";"` turns up a total of two results, both of which look irrelevant. I'm not very knowledgeable with OCaml, but it looks like revised syntax isn't encouraged. Are there any other values that the phrase terminator can take? Should we drop any mentions of support for revised syntax and always require two semicolons?

src/top/utop.el
leungbk: Yeah, I made this in a separate commit to shut up the CI.

src/top/utop.el
leungbk: If I'm not mistaken, the revised syntax was taken from Camlp4/Camlp5, can you confirm? If so, then utop has dropped support for it already, which means that two semicolons should indeed be safe.

bbatsov: I guess @rgrinberg will know best, but I've never seen anything other than `;;` myself.

rgrinberg: We don't need to support the revised syntax at all. It's been dead for a long time already.

leungbk:
OK, I've updated the branch per everyone's suggestions.

bbatsov (approved):
Looks good to me now.

PR #403, opened 2022-10-27
Recommend `opam exec` instead of deprecated `opam config exec`

leungbk:


bbatsov:
This looks reasonable to me. I wonder if we shouldn't just make this the default behavior. I guess it'd be even better to have two commands (one that runs `utop` directly, one that uses `opam` and to just use `locate-dominating-file` (or similar) to figure out if we're in a Dune project.

PR #404, opened 2022-11-03
Shut up Emacs 29's byte-compiler

leungbk:
One of my previous PRs passed CI when I pushed, but failed CI once it got merged. This should fix the issue.


bbatsov:
Looks good. You might also want to fix the docstring, as it would never pass `checkdoc` validation and I guess we'll add this to the CI at some point.

Issue #261, opened 2018-10-16
utop emacs completion-at-point-functions support?


ghost:
Hi. 
Is it possible to add ``completion-at-point-functions`` for utop.el?
I don't want a completion list displayed everytime when I try to complete something, when I don't have company-mode enabled. 
You can use any backend when you have a ``completion-at-point-functions`` (e.g. ivy and company).
I think a ``completion-at-point-functions`` support will be convinent.
Thanks.

rgrinberg:
Sure, feel free to add them but AFAIK the main issue is that you can't support `completion-at-point` asynchronously.

PR #406, opened 2022-12-01
Add completion-at-point implementation ocaml-community/utop#261

j-shilling:
After poking around for ways to use my normal `completion-at-point` workflow in the utop buffer, I came accross issue #261, which is a feature request to provide a completion-at-point-fuction.

This is an implementation I came up with that works for me and I wanted to share it here in case anyone else found it useful or had any feedback.

### DIFF
--- a/src/top/utop.el
+++ b/src/top/utop.el
@@ -101,6 +101,16 @@ This hook is only run if exiting actually kills the buffer."
   :type 'boolean
   :safe 'booleanp)
 
+(defcustom utop-capf-wait-interval 0.01
+  "Length of time to wait when polling for completion candidates."
+  :type 'float
+  :safe 'floatp)
+
+(defcustom utop-capf-max-wait-time 0.1
+  "Maximum time to wait before giving up on completion."
+  :type 'float
+  :safe 'floatp)
+
 (defface utop-prompt
   '((((background dark)) (:foreground "Cyan1"))
     (((background light)) (:foreground "blue")))
@@ -156,6 +166,9 @@ This hook is only run if exiting actually kills the buffer."
 
 (defvar-local utop-completion nil
   "Current completion.")
+
+(defvar-local utop-capf-completion-candidates nil
+  "Current completion when using capf.")
 
 (defvar-local utop-completion-prefixes nil
   "Prefixes for current completion.")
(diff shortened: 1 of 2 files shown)


rgrinberg:
@bbatsov could you review?

src/top/utop.el
```diff
@@ -595,21 +615,43 @@ it is started."
...
                (with-current-buffer utop-complete-buffer
                  (with-output-to-temp-buffer "*Completions*"
                    (display-completion-list (nreverse utop-completion))))
-             (minibuffer-hide-completions))))
+             (minibuffer-hide-completions)))))
        (setq utop-completion nil)))))
 
+(defun utop-completion-at-point ()
+  "Complete thing at point."
+  (when utop-capf-enable
```
bbatsov, line 639: I think this guard here is probably redundant, as you guard for this several places and it's unlikely that people will toggle it off after they loaded the mode.

src/top/utop.el
```diff
@@ -595,21 +615,43 @@ it is started."
...
+    (setq utop-capf-completion nil)
```
bbatsov, line 640: Seems this name is misleading as that's actually the list of completion candidates, right?

bbatsov:
Overall the code looks good to me. In general I think this should be enabled by default - probably unconditionally, given it's the more generic method of completion in Emacs. I see very little point to have to enable manually a `capf` function. Given how basic the company-backend is, probably at some point it can go away, as company has a capf backend anyways.

bbatsov:
While you're at this - I'd also suggest incorporating those doc improvements proposed a while ago https://github.com/ocaml-community/utop/pull/326

j-shilling:
> Overall the code looks good to me. In general I think this should be enabled by default - probably unconditionally, given it's the more generic method of completion in Emacs. I see very little point to have to enable manually a `capf` function. Given how basic the company-backend is, probably at some point it can go away, as company has a capf backend anyways.

As a temporary fix, I added a dispatch function that can be bound to TAB that either calls `completion-at-point` or starts the company workflow. I haven't used company in a long time, but is there any more to this than just removing the current backend and adding `company-capf` backend instead in `utop-mode`?

If it's that simple, I can do that here and then merge in #326 with changes that explain which backend to use. It might also be worth removing the `complete` command and renaming `company-complete` along with removing the `complete-word` response. I'm very new to OCAML, so I might not be the right guy to do that.

bbatsov:
> I haven't used company in a long time, but is there any more to this than just removing the current backend and adding company-capf backend instead in utop-mode?

You don't really need to do anything in utop-mode for `company-capf` to work, provided it registers some completions-at-point function. The purpose of the special company backends is to provide some enhanced functionality, more metadata, etc. Obviously that wasn't the case with the existing backend, that's why I think it becomes kind of redundant. 

> It might also be worth removing the complete command and renaming company-complete along with removing the complete-word response.

Yeah, I'm not sure why it was made company-specific in utop itself. I guess the author of the functionality was a company user and this was before capf became popular (and company alternatives like corfu).

bbatsov:
@rgrinberg @emillon Seems we forgot about this PR, which is good to merge IMO. (especially given that company-mode as not as popular as it used to be)

bbatsov:
Thanks!

PR #409, opened 2022-12-31
use M-down instead of M-tab

sanette:
Both Alt-TAB and Alt-Down already work in utop. But in many OSes, Alt-Tab is reserved for switching windows. Moreover, once your fingers are on LEFT/RIGHT, I think having to move to TAB is not practical. So I think that it makes more sense to advertise ALT-down in the README file.


emillon:
Thanks

PR #418, opened 2023-02-23
Qualify () constructor

### LINKED ISSUE #417: Redefining constructor () with an argument breaks everything
```ocaml
utop # type t = () of unit
Line 1, characters 0-19:
Warning 65 [redefining-unit]: This type declaration is defining a new '()' constructor
which shadows the existing one.
Hint: Did you mean 'type t = unit'?
type t = () of unit
utop # let a = 0;;
Error: The constructor () expects 1 argument(s),
       but is applied here to 0 argument(s)
```

As far as I can tell every subsequent command fails with the same error

Edit: it doesn't happen in the `ocaml` toplevel

emillon:
We are building a hidden expression that contains "()" but it is not
qualified. So it will pick the constructor in scope. This can cause
problems if `()` has been redefined. The correct fix is to qualify it as
part of the `Unit` module.

(additionally, this removes an unused ident)

Fixes #417

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,7 +3,9 @@
 
 * Mark `prompt_continue`, `prompt_comment`, `smart_accept`, `new_prompt_hooks`,
   `at_new_prompt` as deprecated (they have been documented as such since 2012
-  and most of them are ignored) (#xxx @emillon)
+  and most of them are ignored) (#415, @emillon)
+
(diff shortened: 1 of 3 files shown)


Octachron:
An issue with this change just appeared: the `core` library redefine the `Unit` module but without re-exporting `()` thus:
```
open Core;;
1 + 1;;
```
fails with
```
Error: Unbound constructor Unit.()
```

PR #431, opened 2023-06-02
XDG Base Directory Compliance

Skyb0rg007:
This PR adds support for the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) by moving the `~/.utop-history` and `~/.utoprc` files to `$XDG_CACHE_HOME/utop-history` and `$XDG_CONFIG_HOME/utop/utoprc` respectively.
Both of these changes still support fallbacks to their original locations to not introduce a strict backwards incompatibility.

While this history file should really be saved as `$XDG_STATE_HOME/utop-history`, the XDG state directory is currently unsupported in the current version of lambda-term. I have filed a PR to that project [here](https://github.com/ocaml-community/lambda-term/pull/113) to add the functionality for downstream use here.

This PR improves on #362 by utilizing the existing dependency on [lambda-term](https://github.com/ocaml-community/lambda-term), and also updating the location of the `utoprc` file location.

### DIFF
--- a/src/lib/uTop_private.ml
+++ b/src/lib/uTop_private.ml
@@ -9,6 +9,27 @@
 
 open Lwt_react
 
+module Default_paths = struct
+  let ( / ) = Filename.concat
+  let xdg = Xdg.create ~env:Sys.getenv_opt ()
+
+  let resolve ~legacy ~filename =
+    if Sys.file_exists legacy then
+      legacy
+    else
+      filename
+
+  let history_file_name =
+    resolve
+      ~legacy:(LTerm_resources.home / ".utop-history")
+      ~filename:(Xdg.state_dir xdg / "utop-history")
+
+  let config_file_name =
+    resolve
+      ~legacy:(LTerm_resources.home / ".utoprc")
+      ~filename:(Xdg.config_dir xdg / "utoprc")
+end
+
 let size, set_size =
   let ev, set_size = E.create () in
   let init = S.const { LTerm_geom.rows = 25; LTerm_geom.cols = 80 } in
@@ -27,7 +48,7 @@ let ui, set_ui = S.create Console
 
 let error_style = ref LTerm_style.none
 
-(* Config from ~/.utoprc *)
+(* Config from $XDG_CONFIG_HOME/utop/utoprc *)
 let autoload = ref true
 
 let margin_function, set_margin_function =
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,9 @@
 2.13.0 (unreleased)
 -------------------
 
+* Load files from XDG directories (the legacy paths still work). (#431,
+  @Skyb0rg007)
+
 * Remove deprecated values `prompt_continue`, `prompt_comment`, `smart_accept`,
   `new_prompt_hooks`, `at_new_prompt` (#..., @emillon)
 
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ Customization
 ### Colors
 
 To add colors to utop, copy one of the files `utoprc-dark` or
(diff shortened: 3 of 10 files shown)


emillon:
Thanks. I suppose it's fine since we're already depending on lambda-term. But since #362, there's now [`xdg`](https://ocaml.org/p/xdg/latest/doc/Xdg/index.html) which is developed as part of dune. I think it should support the state directory. Have you considered using that directly?

Skyb0rg007:
While I did consider it (I made a PR to add the state dir), it is a very minimal library that is only meant to be used on Unix-like systems. The `xdg` module doesn't do any special handling for Windows, nor does it detect the legacy file locations, both of which I think are important to handle for a cross-platform program.

Skyb0rg007:
While I did consider using `xdg`, it does not currently have support for legacy fallbacks.
However that feature is always possible to implement inside utop itself, rather than via `lambda-term`.
I'll see if I can update to use that instead, since it is more tailored for the intended use-case (and because my PR to `xdg` was accepted, but my PR to `lambda-term` is still waiting for a response).

src/lib/uTop_private.ml
```diff
@@ -9,6 +9,28 @@
...
+  let xdg = Xdg.create ~env:Sys.getenv_opt ()
+
+  (* Fallback to legacy filename *)
+  let resolve legacy fn =
+    if not (Sys.file_exists fn) && Sys.file_exists legacy then
+      legacy
+    else
+      fn
+
+  let history_file_name =
```
emillon, line 23: can this, and `config_file_name`, be functions instead, to avoid IO at module load time?

src/lib/uTop_private.ml
```diff
@@ -9,6 +9,28 @@
 
 open Lwt_react
 
+module DefaultPaths = struct
+  let ( / ) = Filename.concat
+  let xdg = Xdg.create ~env:Sys.getenv_opt ()
+
+  (* Fallback to legacy filename *)
+  let resolve legacy fn =
+    if not (Sys.file_exists fn) && Sys.file_exists legacy then
```
emillon, line 18: I think we can simplify a bit the logic here to "if legacy exists, use it. otherwise, use new path". We'll add a deprecation warning when the legacy path is loaded in a couple releases.

src/lib/uTop_private.ml
```diff
@@ -9,6 +9,28 @@
 
 open Lwt_react
 
+module DefaultPaths = struct
+  let ( / ) = Filename.concat
+  let xdg = Xdg.create ~env:Sys.getenv_opt ()
+
+  (* Fallback to legacy filename *)
+  let resolve legacy fn =
```
emillon, line 17: Can you use named arguments here?

emillon:
also, please add the dependency in the opam file as well.

src/lib/uTop_private.ml
Skyb0rg007: Is `DefaultPaths` idiomatic OCaml, or should the module name be `Default_Paths`? I realize I may make some unidiomatic decisions, since I'm more used to SML.

Skyb0rg007:
The history file should live in `~/.local/state`, but unfortunately [the PR to `xdg`](https://github.com/ocaml/dune/pull/7843) the adds the feature wasn't included in the latest release. I pushed a workaround that does the calculation inline, and can be removed once `Xdg.state_dir` is available.

emillon:
This will get released in dune 3.9.0 which should be out in a bit less than 2 weeks. It's unlikely that there'll be a utop release in the meantime so we can just wait for it to be available.

emillon:
I rebased and used the function from xdg now it's available.

Skyb0rg007:
Thanks for all your work giving feedback!

Issue #433, opened 2023-06-09
utop -stdin crashes


tuohy:
Currently, using the -stdin command line argument results in a fatal exception:

```
$ utop -stdin
Fatal error: exception Sys_error("Is a directory")
```

I have created a [PR](https://github.com/ocaml-community/utop/pull/434) of a clunky workaround that at least demonstrates the underlying issue.

emillon:
Fixed by #434

PR #434, opened 2023-06-10
Fix behavior of utop -stdin

tuohy:
See issue #433 for how the problem showed itself.

In 4.14 the behavior of Toploop.use_silently changed. Previously, the empty string would be interpreted as a request for Stdin. Now File and Stdin are different constructors of Toploop.input.

This PR adds logic to the wrapper already used to use the File constructor to select Stdin when the filename passed is the empty string.

NB this means that `utop -stdin` works but `utop ""` has the same effect, which AFAICT was the behavior pre-4.14 but may not be desirable.


emillon:
Thanks

PR #439, opened 2023-06-21
expunge: move system code to dune

emillon:
The logic to create utop from utop-full is as follows:

- use ocamlobjinfo on utop-full to determine which modules are used
- use ocamlobjinfo on ocamlcommon, ocamlbytecomp, and ocamltoplevel to
  determine which modules to exclude (a priori)
- compute the set used - (exclude - {Topmain,Toploop,Topdirs})
- call expunge with this set of modules

Previously, this was all done in expunge/expunge.ml, meaning core logic
+ process handling.
What this PR does is offload the process handling to dune: calling
ocamlobjinfo and expunge is done instead in dune rules, and
expunge/modules.ml is just about reading sets of module names and
outputing the result.

Besides simplicity and portability, the other advantage is that the
intermediate modules.txt can be inspected directly.


src/top/dune
```diff
@@ -5,13 +5,34 @@
...
+ (with-stdout-to info-ocamltoplevel.txt
+  (run %{bin:ocamlobjinfo} %{lib:compiler-libs.bytecomp:ocamlbytecomp.cma})))
+
+(rule
+ (with-stdout-to info-utop.txt
+  (run %{bin:ocamlobjinfo} %{dep:utop.bc})))
+
+(rule
+ (with-stdout-to modules.txt
+ (run
```
rgrinberg, line 25: missing indentation?

rgrinberg (approved):
LGTM.

Might be worth testing it with sandboxing to see if you got the deps right.

emillon:
this works with sandboxing, and I confirmed (after fixing a typo) that it produces the same binary,

Issue #334, opened 2020-08-06
crash with Fatal error: exception Zed_string.Invalid


jmid:
I use utop in combination with QCheck - for randomized property-based testing.

However, sometimes when I inspect or illustrate QCheck's ability to produce random strings I encounter a crash.

To reproduce:
```ocaml
$ utop
───────────────────────────────┬─────────────────────────────────────────────────────────────┬───────────────────────────────
                               │ Welcome to utop version 2.4.3 (using OCaml version 4.09.0)! │                               
                               └─────────────────────────────────────────────────────────────┘                               
Findlib has been successfully loaded. Additional directives:
  #require "package";;      to load a package
  #list;;                   to list the available packages
  #camlp4o;;                to load camlp4 (standard syntax)
  #camlp4r;;                to load camlp4 (revised syntax)
  #predicates "p,q,...";;   to set these predicates
  Topfind.reset();;         to force that packages will be reloaded
  #thread;;                 to enable threads


Type #utop_help for help about using utop.

─( 13:18:23 )─< command 0 >───────────────────────────────────────────────────────────────────────────────────{ counter: 0 }─
utop # #require "qcheck";;
─( 13:18:23 )─< command 1 >───────────────────────────────────────────────────────────────────────────────────{ counter: 0 }─
utop # open QCheck;;
─( 13:18:30 )─< command 2 >───────────────────────────────────────────────────────────────────────────────────{ counter: 0 }─
utop # Random.init 12; Gen.generate1 ~rand:(Random.get_state ()) Gen.small_string;;
Fatal error: exception Zed_string.Invalid("at position 157: individual combining marks encountered", "- : string =\n\"\195\134x`Ihq\196\178$&\\031\\015\\026k*\\031\194\185\\t\194\179x~jm\195\136i\195\185\195\175\195\166\195\137/\195\156\\025(\194\168\\\\\194\146\195\160^\195\181\194\131[\195\155\195\128\195\161\194\186(5\\021\194\175\202\167\195\180\195\148\195\158\195\1551Y_/E\194\164\194\163\194\191`\195\160h\194\166\195\174\195\153#\194\144\195\188\195\164rG\\026\194\160\194\173\195\152(\194\176\194\158\214\149\\029\194\168G\194\150\\030\194\141\"\n")
```

Note how I explicitly set the random seed to `12` to attempt something reproducable.
[...]

PR #442, opened 2023-06-23
Check that `Zed.next_error` index value is within the string length

tmattio:
Fixes https://github.com/ocaml-community/zed/issues/57


tmattio:
cc @emillon

emillon:
thanks. I'll extract a test harness so that this and #334 can be tested properly

Issue #448, opened 2023-07-10
Escaped backslash at end of string parsed incorrectly


tuohy:
The parser incorrectly associates the second backslash as escaping the quotation. With a highlight setting for strings the problem is clear in this:
![lexbug](https://github.com/ocaml-community/utop/assets/4421874/55575317-cee1-4b3f-ba25-91b60986c3e0)

PR #449, opened 2023-07-10
Parse escaped escapes before escaped quotes

tuohy:
Corrects behavior described in #448 by parsing escaped backslashes before other string contents.
lexbugfixed


kandu:
Thank you, @tuohy :thumbsup:

PR #462, opened 2023-09-10
Bring utop(1) man page up to date with README

gvrooyen:
Here I tried to bring the utop(1) man page a bit more up to date with the README since it contained some information like key bindings that no longer seemed correct. I also included information about Vi mode and Emacs integration. The man pages should be pretty tight and not include things like installation instructions – I rather added a reference to the README itself.

### DIFF
--- a/man/utop.1
+++ b/man/utop.1
@@ -1,11 +1,11 @@
 \" utop.1
 \" ------
-\" Copyright : (c) 2011, Jeremie Dimino <jeremie@dimino.org>
+\" Copyright : (c) 2023, Jeremie Dimino <jeremie@dimino.org>
 \" Licence   : BSD3
 \"
 \" This file is a part of utop.
 
-.TH UTOP 1 "August 2011"
+.TH UTOP 1 "September 2023"
 
 .SH NAME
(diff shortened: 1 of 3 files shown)


man/utop.1
```diff
@@ -1,11 +1,11 @@
 \" utop.1
 \" ------
-\" Copyright : (c) 2011, Jeremie Dimino <jeremie@dimino.org>
+\" Copyright : (c) 2023, Jeremie Dimino <jeremie@dimino.org>
```
emillon, line 3: I don't think it's necessary to change these dates so I'd leave that change out

man/utop.1
```diff
@@ -1,11 +1,11 @@
...
 \" Licence   : BSD3
 \"
 \" This file is a part of utop.
 
-.TH UTOP 1 "August 2011"
+.TH UTOP 1 "September 2023"
```
emillon, line 8:
```suggestion
.TH UTOP 1
```

No need to specify a date here.

emillon (changes requested):

emillon (approved):

emillon:
Thanks!

Issue #412, opened 2023-01-24
Create implicits does not work in Emacs mode


ysalmon:
When using UTop in Emacs, the ``-implicit-bindings`` option seems to have no effect.

On [discuss.ocaml.org](https://discuss.ocaml.org/t/utop-create-implicits-not-working-with-emacs/11229/2), it was pointed out that uTop uses a different function when processing in emacs mode, that does not refer to ``get_create_implicits`` : https://github.com/ocaml-community/utop/blob/595002e6f07e6a3c6abc6e94a1b2448006115f1b/src/lib/uTop_main.ml#L1078

But they do not know why.

PR #465, opened 2023-10-05
Fix: Make emacs mode respect implicit bindings

bencef:


bencef:
Propsed fix for #412

Octachron:
The code is correct and I agree that it is better to align the emacs mode with the standard mode.

PR #468, opened 2023-11-11
Fix completion of symbols in modules

### LINKED ISSUE #455: Completion broken when running utop from emacs
Typing `Lis<TAB>` correctly completes into either `List` or `ListLabels` but when typing `List.<TAB>`  a message "No match" appears in the minibuffer.

```
 utop -version
The universal toplevel for OCaml, version 2.13.1, compiled for OCaml version 5.0.0

ocamlc -version
5.0.0

emacs --version
GNU Emacs 29.1.50
```

steffenhaug:
Fixes #455 

The reason you don't get a completion when typing a symbol in a module, is because the range from `utop-prompt-max` to `(point)` does not match the actual prefix of the completion returned from `utop`. The fix is to set the `start` and `end` positions of the completion corresponding to the position of the prefix `utop` is actually completing.

Before, there was a fix that attempted to fix the same thing with some special logic in the processing of the lines returned from `utop`, but using the `cadr` of the split string will not work with more than one level of modules, and even then the range of the completion was not set correctly, leading to the issue. With this fix, we can complete things like `List.<TAB>`, and even `Base.Float.O.<TAB>`.

There might be a better way to determine the completion prefix; I'm not experienced enough with OCaml to know if splitting by `.` covers finding the prefix for everything that `utop` can autocomplete, but at least it is a start.

I'm also not the most experienced elisp hacker, so if the code is in bad style or there is something I have overlooked,  feel free to make suggestions as to how I can improve the fix.

### DIFF
--- a/src/top/utop.el
+++ b/src/top/utop.el
@@ -600,16 +600,7 @@ it is started."
        (setq utop-completion nil))
       ;; A new possible completion
       ("completion"
-       (catch 'done
-         (dolist (prefix utop-completion-prefixes)
-           ;; We need to handle specially prefixes like "List.m" as
-           ;; the responses from utop don't include the module prefix.
-           (let ((prefix (if (string-match-p "\\." prefix)
-                             (cadr (split-string prefix "\\."))
-                           prefix)))
-             (when (string-prefix-p prefix argument)
-               (push argument utop-completion)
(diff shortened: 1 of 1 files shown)


Octachron:
Superseded by #497 .

PR #470, opened 2023-11-13
Support OCaml 5.2

### LINKED ISSUE #466: OCaml 5.2 support
At the moment, `utop` is not compatible with OCaml 5.2.0 (or rather, with the current version of `trunk`, since 5.2.0 has not been released  yet).

To support this, it is necessary to update `uTop_compat.ml` to account for the API changes in the compiler libs.
This is usually straightforward (one parameter is added, a type becomes abstract, etc) but requires good familiarity with OCaml and build tools (`dune`, `opam` and `cppo`).

I can provide help with this.

leostera:
Hi folks! I saw Sabine's tweet and I figured I'd give this a shot. There's a few things that I'd like to upstream to the compiler (like Ast_helpers to create function params), but otherwise this seems to be working.

Screenshot 2023-11-13 at 18 05 01

Happy to amend anything here, just figured I'd open the PR early to get feedback / directions.

Ideally would close #466

### DIFF
--- a/src/lib/uTop.ml
+++ b/src/lib/uTop.ml
@@ -288,9 +288,8 @@ let parse_default parse str eos_is_error =
       | Syntaxerr.Ill_formed_ast (loc, s) ->
           Error ([mkloc loc],
                  Printf.sprintf "Error: broken invariant in parsetree: %s" s)
-      | Syntaxerr.Invalid_package_type (loc, s) ->
-          Error ([mkloc loc],
-                 Printf.sprintf "Invalid package type: %s" s)
+      | Syntaxerr.Invalid_package_type (loc, err) ->
+          Error ([mkloc loc], UTop_compat.invalid_package_error_to_string err)
 #if OCAML_VERSION >= (5, 0, 0)
       | Syntaxerr.Removed_string_set loc ->
           Error ([mkloc loc],
@@ -358,11 +357,12 @@ let check_phrase phrase =
           let open Ast_helper in
           with_default_loc loc
             (fun () ->
-               Str.eval
-                 (Exp.fun_ Nolabel None (Pat.construct unit None)
-                   (Exp.letmodule (with_loc loc (Some "_"))
+              let punit = (Pat.construct unit None) in
+              let body = (Exp.letmodule ~loc:loc
+                      (with_loc loc (Some "_"))
                       (Mod.structure (item :: items))
-                      (Exp.construct unit None))))
+                      (Exp.construct unit None)) in
+              Str.eval (UTop_compat.Exp.fun_ ~loc punit body))
         in
         let check_phrase = Ptop_def [top_def] in
         try
@@ -828,7 +828,8 @@ let () =
    | Compiler-libs re-exports                                        |
    +-----------------------------------------------------------------+ *)
 
-let get_load_path () = Load_path.get_paths ()
+let get_load_path = UTop_compat.get_load_path
+
 let set_load_path = UTop_compat.set_load_path
 
 module Private = struct
--- a/src/lib/uTop_compat.ml
+++ b/src/lib/uTop_compat.ml
@@ -19,11 +19,21 @@ let toploop_all_directive_names () =
   Hashtbl.fold (fun dir _ acc -> dir::acc) Toploop.directive_table []
 #endif
 
-let set_load_path path =
-#if OCAML_VERSION >= (5, 0, 0)
-  Load_path.init path ~auto_include:Load_path.no_auto_include
+let get_load_path () =
+#if OCAML_VERSION >= (5, 2, 0)
+  let {Load_path.visible; hidden} = Load_path.get_paths () in
+  visible @ hidden
 #else
-  Load_path.init path
+  Load_path.get_paths ()
+#endif
+
+let set_load_path visible =
+#if OCAML_VERSION >= (5, 2, 0)
+  Load_path.init ~auto_include:Load_path.no_auto_include ~visible ~hidden:[]
+#elif OCAML_VERSION >= (5, 0, 0)
+  Load_path.init ~auto_include:Load_path.no_auto_include visible
+#else
+  Load_path.init visible
 #endif
 
 let toploop_use_silently fmt name =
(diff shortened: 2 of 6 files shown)


src/lib/uTop_compat.ml
```diff
@@ -19,11 +19,21 @@ let toploop_all_directive_names () =
...
+  { visible = Load_path.get_load_path (); hidden = [] }
+#endif
+
+let set_load_path {visible; hidden} =
 #if OCAML_VERSION >= (5, 0, 0)
-  Load_path.init path ~auto_include:Load_path.no_auto_include
+  Load_path.init ~auto_include:Load_path.no_auto_include ~visible ~hidden
 #else
-  Load_path.init path
+  Load_path.init (visible @ hidden)
```
leostera, line 36: Probably should just use `visible` here.

ManasJayanth:
Got here too late: My branch is here https://github.com/DiningPhilosophersCo/utop/tree/prometheansacrifice%40ocaml-5-2
Atleast I think I can lend a hand with a code review. Will leave a comment if necessary. This was fun!

emillon:
Thanks for this PR.
I think that this is going in the right direction but the main thing that is left to address is compatibility. This version of utop needs to build and run on OCaml 4.11 to 5.2.0 (that seems large but bear in mind that I've already axed much of the compatibility so it's a lot easier than it used to be).
To have a system where it is easy to test compat code, I recommend the following set up:
- put this in `dune-workspace.multi` (do not commit this file to the repository)
```
(lang dune 1.0)
(context (opam (switch utop-411)))
(context (opam (switch utop-412)))
(context (opam (switch utop-413)))
(context (opam (switch utop-414)))
(context (opam (switch utop-500)))
(context (opam (switch utop-510)))
(context (opam (switch utop-520)))
```
- create N *global* switches named like the above and based on the corresponding ocaml compiler version, and install the dev-deps in them. You'll never activate these switches directly

    $ opam sw create utop-411 ocaml-base-compiler.4.11.1
    $ opam install --switch utop-411 utop --deps-only --with-test
    ...
    $ opam sw create utop-510 ocaml-base-compiler.5.1.0
    $ opam install --switch utop-510 utop --deps-only --with-test
    $ opam sw create utop-520 ocaml-variants.5.2.0+trunk
    $ opam install --switch utop-520 utop --deps-only --with-test

(note the `--switch ...` flags passed to `install`, and the fact we used `ocaml-base-compiler` everywhere but `ocaml-variants` in the `5.2.0` case)

- then instead of the usual `dune build`, use `dune build --workspace dune-workspace.multi`. This is going to make a dune build that builds using all the different switches. Error messages will carry the name of the opam switch that failed, so you can use that to know what went wrong.

This is the main trick and the most important part to get a green build on all supported versions.

Now, regarding architecture, utop used to have many `#if`s sprinkled all around the code base. I moved most of them to `uTop_compat.ml`. Please aim to put new compat code in there but I appreciate that it's not always possible.

I'll leave the rest of my comments as a review. Thanks again.

src/lib/uTop.ml
```diff
@@ -288,9 +288,32 @@ let parse_default parse str eos_is_error =
       | Syntaxerr.Ill_formed_ast (loc, s) ->
           Error ([mkloc loc],
                  Printf.sprintf "Error: broken invariant in parsetree: %s" s)
-      | Syntaxerr.Invalid_package_type (loc, s) ->
-          Error ([mkloc loc],
-                 Printf.sprintf "Invalid package type: %s" s)
+      | Syntaxerr.Invalid_package_type (loc, err) ->
+#if OCAML_VERSION >= (5, 2, 0)
```
emillon, line 292: Can you move that block to `UTop_compat` as a `invalid_package_error_to_string`? This module has no `mli` file so it can take a string in <5.2.0 and the variant in >=5.2.0.

src/lib/uTop_main.ml
```diff
@@ -565,9 +571,20 @@ let rewrite_rules = [
...
       let punit = Pat.construct (with_loc loc (Longident.Lident "()")) None in
       with_default_loc loc (fun () ->
+#if OCAML_VERSION >= (5, 2, 0)
+        let open Parsetree in
+        let args = [{
+          pparam_loc=loc;
+          pparam_desc=Pparam_val (Nolabel, None, punit) 
+        }]
+        in
+        let f = Exp.function_ args None (Pfunction_body e) in
```
emillon, line 581: This can use the `UTop_compat.fun_` helper described before.

src/lib/uTop_main.ml
```diff
@@ -336,8 +336,14 @@ end = struct
   let scan_cmis =
     let new_cmis = ref [] in
     let default_load = !Persistent_env.Persistent_signature.load in
-    let load ~unit_name =
+
```
emillon, line 339: I think that this part can be made into a `add_cmi_hook : (cmi_infos -> unit) -> unit` which will also improve `scan_cmis`.

src/lib/uTop_main.ml
```diff
@@ -582,7 +599,12 @@ let rule_path rule =
       let env = !Toploop.toplevel_env in
       let path =
         match Env.find_type_by_name rule.type_to_rewrite env with
-        | path, { Types.type_kind     = Types.Type_abstract
+        | path, {
+#if OCAML_VERSION >= (5, 2, 0)
+                  Types.type_kind     = Types.(Type_abstract Definition)
```
emillon, line 604: This can be moved to `UTop_compat.type_abstract_definition`

src/lib/uTop_main.ml
```diff
@@ -1545,7 +1567,12 @@ let interact ?(search_path=[]) ?(build_dir="_build") ~unit ~loc:(fname, lnum, cn
   let search_path = walk build_dir ~init:search_path ~f:(fun dir acc -> dir :: acc) in
   let cmt_fname =
     try
-      Misc.find_in_path_uncap search_path (unit ^ ".cmt")
+#if OCAML_VERSION >= (5, 2, 0)
+      Misc.find_in_path_normalized
```
emillon, line 1571: This can be moved to `UTop_compat.find_in_path_normalized`

src/lib/uTop_main.ml
```diff
@@ -1577,7 +1604,14 @@ let interact ?(search_path=[]) ?(build_dir="_build") ~unit ~loc:(fname, lnum, cn
...
   try
-    List.iter Topdirs.dir_directory (search_path @ cmt_infos.cmt_loadpath);
+    let visible_paths = 
+#if OCAML_VERSION >= (5, 2, 0)
+      cmt_infos.cmt_loadpath.visible
+#else
+      cmt_infos.cmt_loadpath
+#endif
+    in
+    List.iter Topdirs.dir_directory (search_path @ visible_paths);
```
emillon, line 1614: Same here this can be moved to a new function in `UTop_compat`.

src/lib/uTop.mli
```diff
@@ -340,8 +340,9 @@ val split_words : string -> string list
 
  (** {6 compiler-libs reexports} *)
 
-val get_load_path : unit -> string list
-val set_load_path : string list -> unit
+type paths = { visible : string list; hidden : string list; }
+val get_load_path : unit -> paths
```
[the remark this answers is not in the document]
ManasJayanth: I had attempted this PR too. I'm curious - what the right order here? `visible @ hidden` ?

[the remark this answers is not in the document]
ManasJayanth: (thats what I had done)

src/lib/uTop_compat.ml
```diff
@@ -19,11 +19,26 @@ let toploop_all_directive_names () =
...
+
+let set_load_path paths =
+#if OCAML_VERSION >= (5, 2, 0)
+  let { visible; hidden } = paths in
+  Load_path.init ~auto_include:Load_path.no_auto_include ~visible ~hidden
+#elif OCAML_VERSION >= (5, 0, 0)
+  Load_path.init ~auto_include:Load_path.no_auto_include paths.visible
 #else
-  Load_path.init path
+  Load_path.init paths.visible
```
ManasJayanth, line 41: @emillon Would this be`visible @@ hidden`

ManasJayanth: Does utop need a `~hidden` flag like the compiler? Ref: https://github.com/ocaml/ocaml/commit/01178bb60eb337790c597af937ac79419eab241c

src/lib/uTop_main.ml
leostera: Where's `add_cmi_hook` coming from?

leostera:
@emillon thanks for the guidance! 🙏🏼 made most of the changes (except the `add_cmi_hook`, but left you a question there). Let me know how you want to proceed 🙌🏼

src/lib/uTop_main.ml
emillon: Sorry, let me expand what I meant here.
`scan_cmis` does two things:
- it changes `Persistent_env.Persistent_signature.load` so that each time it's called, the result value is added to a `new_cmis` global (even if it has a local lexical scope)
- it returns a function that consumes this list

To support 5.2 we need to alter the first part. My suggestion is to cut the function in two parts:

- part 1 in `UTop_compat`
```ocaml
let add_cmi_hook f =
#if OCAML_VERSION...
  let load ...
#endif
    Option.iter (fun x -> f x.cmi) res
  in
  Persistent_env....load := load
```

- part 2 here:

```ocaml
let scan_cmis =
  let new_cmis = ref [] in
  add_cmi_hook (fun cmi -> new_cmis := cmi :: !new_cmis);
  fun pp ->
    ...
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,44 @@
...
+Some code will be different from one version of OCaml to the next. If you find
+some that does, please abstract it away using the `UTop_compat` module.
+
+For example, the `Load_path.get_paths ()` function has changed recently to
+return a record with shape `{ visible: string list; hidden: string list }`, but
+this function used to return a single `string list`.
+
+Defining this function using pre-processor macros allows us to give the same
+function two different bodies on different version of the language. Note that
+they must have the exact same type signature.
```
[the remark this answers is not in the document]
leostera: Not sure what you want me to do here :P – I'm happy rewriting/removing parts of that file, but I meant to write something that'd lead other folks to an easy time contributing. 

In my case, seeing UTop_compat as having a fixed interface made it easier to modify things in it, because I had a contract to comply to.

I understand there's nuance here, and that whatever signature the UTop_compat has is find as long as it fits the usage, but then I have to also consider and write about this, and for someone getting started this nuance seemed unnecessary.

Agin, happy to rewrite! Just let me know what would make sense to you (or if I missed something, what I misssed) 🙌🏼

emillon:
(I haven't forgotten, I'll get back to you about this during week 1)

src/lib/uTop.mli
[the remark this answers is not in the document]
emillon: updated to include hidden

emillon:
@leostera I rebased this on top of `master`, added a changelog crediting all coauthors.

This is good to go for a release supporting 5.2 but there will probably be things to fix regarding hidden paths, and we'll have to decide if we want to provide a precise API for the load path.

emillon:
Thanks everyone. that was the first time I was looking for external contributors and this was very succesful, despite the delay in merging!

kit-ty-kate:
Would it be possible to get a new release?

pmetzger:
@kit-ty-kate You don't have access to do that on your own? We could give it to you.

emillon:
as I said previously, I'm waiting for 5.2 in ocaml-ci to cut the release to make sure it works there. this should happen after the first alpha, right?

kit-ty-kate:
alpha1 is already available but ocaml-ci hasn't been updated yet. This is tracked at https://github.com/ocurrent/ocaml-ci/issues/917

In the meantime, is there any chance you could spawn an opam switch locally instead? It should take less than 3 minutes

Issue #471, opened 2023-12-05
Emacs doesn't automatically open .ocamlinit in utop


benbellick:
# Summary
Hi! I am trying to use `utop-mode` in my emacs workflow and am encountering the following issue. I would like to do some basic setup in an `.ocamlinit` file to make certain facilities available to me every time I start a new `utop` (e.g. install a printer). 

# Example
Suppose I have a `.ocamlinit` file which looks as follows:
```ocaml
open Utop_test__Lib;;
#install_printer pp_a;;
```
where the first line is just some example module, and the second line is just some example printer being installed (which we will say prints values of type `a`. 
When I start `utop` from the command line via `opam exec -- dune utop . -- -emacs`, `Utop_test__Lib` is indeed opened and `pp_a` is automatically called when values of type `a` are created. 

However, when I run the (nearly identical) command `opam exec -- dune utop . -- -emacs` when invoking `M-x utop`, the `.ocamlinit` file is not loaded. 

Is this the intended behavior? If not, I would be happy to work on it if someone can point me towards where to start :)

benbellick:
Upon further investigation, I now see that the problem is the command `opam exec -- dune utop . -- -emacs` is executed from dir in which the currently focused buffer exists. Is there currently some way I can tell utop to default execute from the project root dir?

PR #472, opened 2023-12-13
Customize starting directory for utop

benbellick:
Hi! I opened up #471 . I use `utop-mode` but wanted to automatically load in my `.ocamlinit` file at the root level of my project without having to pass it in explicitly using the `-init` flag. My solution is to enable a customization to start utop from the project root dir. 

Now when you set the variable `utop-starting-directory` (possibly via customize) to `'project-root`, `project.el` is used to find the current buffer's project root. Then `utop` is called from that directory. If the project cannot be found, an error is thrown.

### DIFF
--- a/src/top/utop.el
+++ b/src/top/utop.el
@@ -75,6 +75,18 @@ with Emacs to provide an enhanced environment."
   :type 'string
   :safe 'stringp)
 
+(defcustom utop-starting-directory 'current-directory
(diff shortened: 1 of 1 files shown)


benbellick:
This actually breaks completion when the directory is set to the project root. Will investigate more but if you have any ideas what might cause this issue, please let me know! Thanks

Issue #481, opened 2024-03-05
Colors not loaded from `~/.config/utop/utoprc`


SinTan1729:
The README says that to enable colors, one needs to edit `~/.config/utop/utoprc`. I used the `utoprc-dark` template. But it doesn't work. It does work however when copied to `~/.config/utoprc`.

Octachron:
The location of those files will be more regular (at .config/utop/utoprc) starting with utop 2.16.0 .

PR #475, opened 2024-01-15
Update README.md

shantanu-sardesai:
Hi, I started learning OCaml today and was installing `utop` when I noticed a minor inconsistency in between the directions given in the `utoprc-dark` file and the `README` on the location to store the theme information.

The ones in the `utoprc-dark` worked for me, so creating a PR to update the path in the `README`.

Issue #478, opened 2024-01-31
History silently not saved between sessions if parent directory of history file doesn't exist


johnmasson:
UTop was not saving history between sessions on my installation. After a bit of poking around, I discovered that UTop.history_file_name was set to '\~/.local/state/utop-history' and that the '\~/.local/state' directory did not exist. Manually creating the directory allowed UTop to start saving history between sessions.

I couldn't find any mention of the need to create this directory in documentation either for UTop or LTerm, and UTop doesn't print any warnings for this situation. It seems like it would be better to create the directory if it doesn't exist, or at least warn that history will not be saved. 

Platform info:
Ubuntu 22.04.3 LTS on WSL (Windows 11)
OCaml version 5.1.1
UTop and all modules : latest versions from OPAM

PR #484, opened 2024-04-03
Correct paths for utop-history, utoprc: in utop subdirs of XDG base dirs

tuohy:
Issue #481 described failure to load utoprc from its stated location
PR #475 misidentified the issue as an incorrect README

In fact the paths were pointing to files within the XDG base directories themselves rather than in a subdirectory for the utop application.

This patch simply changes $XDG_STATE_HOME/utop-history to $XDG_STATE_HOME/**utop**/utop-history and $XDG_CONFIG_HOME/utoprc to $XDG_CONFIG_HOME/**utop**/utoprc

### DIFF
--- a/src/lib/uTop_main.ml
+++ b/src/lib/uTop_main.ml
@@ -53,7 +53,10 @@ let init_history () =
         return ()
     | Some fn ->
         Lwt.catch
-          (fun () -> LTerm_history.load UTop.history fn)
+          (fun () ->
+            let dn = Filename.dirname fn in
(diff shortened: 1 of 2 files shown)


tuohy:
Added check & creation of history file's parent directory, fixing #478

This was exacerbated with the first patch because XDG_STATE_HOME/utop was extremely unlikely to exist on any system

mmottl:
I believe this should be the proper place since `init.ml` is also in `.config/utop`, and it's already documented like that in the README.

PR #489, opened 2024-09-03
OCaml 5.3 support

anmonteiro:


pmetzger:
Looks at worst harmless in that it only impacts 5.3 and after. However, I think as part of this we should extend the CI to 5.3 since the CI didn't actually check that.

src/lib/uTop_compat.ml
```diff
@@ -67,7 +67,28 @@ let rec is_persistent_path = function
 #endif
 
 let invalid_package_error_to_string err =
-#if OCAML_VERSION >= (5, 2, 0)
+#if OCAML_VERSION >= (5,3,0)
```
kit-ty-kate, line 70: this section can be nested in the original one. Only a two lines are actually different

Octachron: I would suggest  to define `inline_code` compatibility helper rather than deduplicate the full printer.

anmonteiro: I thought only the newest cppo had nesting support, so I avoided doing it. Am I wrong?

anmonteiro:
Thanks for the helpful code review. The patch is way more minimal now.

pmetzger:
Okay, so what's next?

anmonteiro:
I'm happy to lend some time if utop needs more maintainers

Octachron (approved):
The changes look good to me.

anmonteiro:
thanks for merging. would be nice to put together an opam release of utop compatible with OCaml 5.3

Octachron:
I will cut a release after the OCaml 5.3.0~beta1 and the 5.2.1 release (aka at the end of this week or next week).

PR #497, opened 2025-01-23
Emacs completion: Include module prefix in completion candidates

### LINKED ISSUE #455: Completion broken when running utop from emacs

juergenhoetzel:
Fixes a mismatch of the completion START position in the Emacs buffer and the candidatates that didn't include the expected module-prefix.

Fixes #455

### DIFF
--- a/src/top/utop.el
+++ b/src/top/utop.el
@@ -604,11 +604,10 @@ it is started."
          (dolist (prefix utop-completion-prefixes)
            ;; We need to handle specially prefixes like "List.m" as
            ;; the responses from utop don't include the module prefix.
-           (let ((prefix (if (string-match-p "\\." prefix)
-                             (cadr (split-string prefix "\\."))
-                           prefix)))
-             (when (string-prefix-p prefix argument)
-               (push argument utop-completion)
+           (let* ((pos (string-match  "[^.]*$" prefix))
+                  (module-prefix (substring prefix 0 pos)))
+             (when (string-prefix-p (substring prefix pos) argument)
+               (push (concat module-prefix argument) utop-completion)
                (throw 'done t))))))
       ;; End of completion
       ("completion-stop"


juergenhoetzel:
@bbatsov You contributed the original code: Could you take a look?

src/top/utop.el
```diff
@@ -604,12 +604,10 @@ it is started."
...
            ;; We need to handle specially prefixes like "List.m" as
            ;; the responses from utop don't include the module prefix.
-           (let ((prefix (if (string-match-p "\\." prefix)
-                             (cadr (split-string prefix "\\."))
-                           prefix)))
-             (when (string-prefix-p prefix argument)
-               (push argument utop-completion)
-               (throw 'done t))))))
+           (when-let* ((pos (string-match  "[^.]*$" prefix))
+                       ((string-prefix-p (substring prefix pos) argument)))
```
bbatsov, line 608: I don't like much pushing conditional checks without bindings to `when-let`, as this makes a bit harder to understand what the code does.

src/top/utop.el
```diff
@@ -604,12 +604,10 @@ it is started."
...
+             (push (concat (substring prefix 0 pos) argument) utop-completion)
```
bbatsov, line 609: Probably a binding for `(substring prefix 0 pos)` will make this bit of code slightly easier to understand.

juergenhoetzel: Thanks for your helpful suggestions. I amended the commit.

bbatsov (approved):
The proposed change seems solid to me!

Octachron:
The code looks correct from afar (I am unfortunately  less familiar with emacs than with the compiler), however I am not been able to reproduce the bug report yet. @juergenhoetzel , do you have a full reproduction case that I could follow to check the bug and the fix in order to move forward with this PR?

juergenhoetzel:
> The code looks correct from afar (I am unfortunately less familiar with emacs than with the compiler), however I am not been able to reproduce the bug report yet. @juergenhoetzel , do you have a full reproduction case that I could follow to check the bug and the fix in order to move forward with this PR?

Of course:

1. Start `utop`: <kbd>M-x utop</kbd>
2. Enter prefix `Filename.co` and press <kbd>tab</kbd>

Result: No completions 

Using this PR, this will result in `Filename.concat`

Octachron:
I finally found the time to test the fix, sorry for the wait ! I have merged the PR and it should be part of the upcomong 2.16.0 release this week.

PR #514, opened 2026-05-01
Add ability to toggle off welcome message

jg791:
I prefer running `utop` as a stripped down interpreter in the shell, which is mostly achievable with the existing configuration options. However, the welcome message tends to take up space and "break" if you resize a terminal window. This PR adds a handy config option, `UTop.set_show_welcome` to disable it (similar to `UTop.set_show_box`), while still defaulting to an "enabled" setting to preserve backwards compatibility. 

The README and man pages are updated accordingly.


jg791:
image



Screenshot of starting `utop` with the following in `~/.config/utop/init.ml`:

```ocaml
#utop_prompt_dummy;;
UTop.set_show_box false;;
UTop.set_show_welcome false;;
```

src/lib/uTop_main.ml
```diff
@@ -1503,10 +1503,10 @@ let main_aux ~initial_env =
       UTop_private.set_size (S.const (LTerm.size term));
       (* Load user data. *)
       Lwt_main.run (Lwt.join [UTop_styles.load (); load_inputrc ()]);
-      (* Display a welcome message. *)
-      Lwt_main.run (welcome term);
       (* Common initialization. *)
       common_init ~initial_env;
+      (* Display a welcome message, if enabled. *)
+      if UTop.get_show_welcome () then Lwt_main.run (welcome term);
```
jg791, line 1509: Moved so that the initialization is loaded before showing the welcome. Otherwise, the instructions in `init.ml` are ignored when deciding if we show the welcome message.

jg791:
@rgrinberg This has been open for a few months now, is there any possibility to get a yay or nay on this? Thank you.

pmetzger:
The idea seems perfectly innocuous to me, for what it's worth.

rgrinberg:
No objections from me. Nor is it necessary to wait for my opinion.

jg791:
So FYI @rgrinberg appears to be automatically tagged as a necessary review (being the codeowner) before I can merge. I do not seem to have the ability to request another review. If there's something else I need to do here or a step I missed, please let me know.

PR #517, opened 2026-07-12
Bring the man pages up to date

bbatsov:
I came across the man pages randomly while looking for the syntax highlighting config and figured the stale bits were worth fixing.

They still documented the pre-XDG paths (`~/.utoprc`), even though the config moved to `$XDG_CONFIG_HOME/utop/utoprc` in #484. I've updated utop(1) and utoprc(5) to document the current locations (noting the legacy fallbacks still work), added the previously undocumented history file to FILES, and pointed to the shipped utoprc-dark/utoprc-light configs.

Along the way I also fixed a few small errors: utop-full(1) claimed the binary is called `utop`, `-safe-string` no longer exists on OCaml >= 5.0, the Emacs section was missing reason-mode and `utop-mode-compat-alist`, plus a couple of typos.

### DIFF
--- a/man/utop-full.1
+++ b/man/utop-full.1
@@ -1,17 +1,17 @@
-\" utop.1
-\" ------
+\" utop-full.1
+\" -----------
 \" Copyright : (c) 2011, Jeremie Dimino <jeremie@dimino.org>
 \" Licence   : BSD3
 \"
 \" This file is a part of utop.
 
-.TH UTOP-FULL 1 "August 2011"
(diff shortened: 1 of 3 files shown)


pmetzger:
The changes look good to me, I don't understand why the CI failed...

jg791:
> The changes look good to me, I don't understand why the CI failed...

It's doing the same to me on my PR here: https://github.com/ocaml-community/utop/pull/514

With no changes it went from succeeding to randomly failing on various builds. Sometimes if I re-run it it succeeds, other times not.

PR #519, opened 2026-07-31
Add OCaml 5.6 support

kit-ty-kate:


Octachron:
Question to @kit-ty-kate  and @anmonteiro would you prefer that I review and merge the 5.6 PR supports on the fly, or would you rather wait for the beta release? The first option is more work but I have the impression that it might be easier for people to use utop with `--dev-repo` pin in this case.

kit-ty-kate:
personally i think it would make more sense to wait until the freeze, in case master is used to do a release and we end up having two releases that have "Add/Fix support for 5.6".

My branch of utop is already in opam-alpha-repository in case someone want an easy way to get it (and i'm guessing @anmonteiro's is in the ocaml-nix overlay)

anmonteiro:
> My branch of utop is already in opam-alpha-repository in case someone want an easy way to get it (and i'm guessing @anmonteiro's is in the ocaml-nix overlay)

this is correct, but I think we can just close mine if this is more complete!

anmonteiro:
@kit-ty-kate this needs another small fix after https://github.com/ocaml/ocaml/pull/15023:

```diff
diff --git a/src/lib/uTop_compat.ml b/src/lib/uTop_compat.ml
--- a/src/lib/uTop_compat.ml
+++ b/src/lib/uTop_compat.ml
@@ -28,7 +28,11 @@ let destruct_ldot p s =
 (* Check whether an identifier is a valid one. *)
 #if OCAML_VERSION >= (5, 3, 0)
 let is_valid_identifier id =
+#if OCAML_VERSION >= (5, 6, 0)
+  Utf8_lexeme.is_valid_identifier id
+#else
   Misc.Utf8_lexeme.is_valid_identifier id
+#endif
 let lax_modname_from_cmi = Unit_info.lax_modname_from_source
 #else
 let is_valid_identifier id =

```

kit-ty-kate:
@anmonteiro thanks, done! I've reduced the fix to a toplevel module alias if the < 5.6 case to make it slightly more generic

